PR #2, opened 2023-10-01
Us17t41

mccodav1:
The README has been updated with a brief rundown of the project structure, where certain things are located, and how pieces interact with each other.

I began to play around with LoginActivity and RegisterActivity to try and see how it could be integrated with the new API at https://github.com/Appyo-Poste/PosteBackend; some changes reflect this intention. Various methods and functions have been documented as I traced the control flow of the program.

At this time, I have a pretty good understanding of how the program runs, how a User logs in or registers, and how Android determines which windows to open, when, and how the Activities correspond to layouts in `res` folder xml files.

More work will need to be done on the API and editing the calls herein before I can begin to see how later layers work.


dzemlin (approved):
All looks good to me. Though admittedly I can't run a test till we can get a fake login system going.

PR #3, opened 2023-10-03
Us9 use case diagram

dzemlin:
Create a Use Case Diagram for Poste


mccodav1:
Thanks for the UCD, David. I might suggest breaking the Use Case Diagram into Scenarios, and then outlining Scenarios (see [here ](https://ibb.co/N3C5ZVq) for an example) The diagram here indicates a user can directly do all these actions, but they're actually broken up by different screens with different actions, so breaking into scenarios seems to best describe how the system is used.

If you look into src/main/res you can see the different layouts -- in Android Studio, you'll see them graphically. You also might base it on the screens Fayez has proposed, as that is the functionality we will seek to implement.

Also, basically everything interacts with the API -- there is no 'authenticator' system except for inside the API. I'm happy to discuss this with anyone interested in a huddle, but might be difficult to discuss textually.

Ideally, I think a user might interact with different described scenarios, and the scenarios themselves interact with the API, which may/may not call the authenticate functionality (only used for logging in, I believe, which gatekeeps the rest of the processes, except registration)

The oddball here is 'sharing content' which really occurs outside the system, and then the OS interacts with Poste itself. Having system boundaries might help describe this behavior

dzemlin:
Applied revision based on the above concerns.

mccodav1 (approved):
I think this better captures the system, nice work

PR #6, opened 2023-10-12
Add HTTP utils and new RegisterActivity

mccodav1:
This change introduces new HTTP utility methods to make HTTP requests easier. It also introduces a new RegisterActivity that allows users to register for the app.

Note that this includes mandatory changes to the AndroidManifest.xml as well as both the project and app build.gradle files.


atreichl (approved):
looks good to me and the calls are a lot simpler then what the code did before.

PR #8, opened 2023-10-20
Refactor login activity

fmehdad97:
This commit refactors the login activity:
- activity_login.xml reflects the layout that was approved by the sponser earlier. 
- Completely refactors the functionality of the login activity 
   - Interacts with the new API developed
   - Uses the token functionality to updateFoldersAndPosts
- Also further updates data models that were created by @mccodav1, specifically to add or modify a few members and to fix folder and posts being replicated on updates.
- Created LoginRequest object, to send to API
- Added additional drawable files of updated logo and icons


app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
-    private String password;
-    private User loginUser;
+    private EditText usernameField, passwordField;
 
     /**
-     * Called when the activity is created
-     *
-     * @param savedInstanceState A bundle containing the saved instance state
+     * function that runs on creation of the activity
+     * @param savedInstanceState
```
mccodav1: I appreciate the docstrings, thank you. It might be helpful to describe the params throughout; this way, the docstring tooltip will explain what params are expected to be

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
         setContentView(R.layout.activity_login);
 
-        // Prep vars
-        usernameField = findViewById(R.id.LoginEmailEntertxt);
-        passwordField = findViewById(R.id.LoginPassword);
-        rememberMeSwitch = findViewById(R.id.login_remember_me_switch);
-
-        //Email = findViewById(R.poste_item_id.LoginEmailEntertxt);
+        TextView hyperlinkTextView = findViewById(R.id.hyperlinkTextViewToRegister);
+        usernameField = findViewById(R.id.editTextTextEmailAddress);
```
mccodav1, line 51: Some of the var names seem a little redundant; perhaps a typo? Might help clarity to rename, but not a blocker.

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
+        /**
+         * adds a hyperlink to redirect user to the login page
+         */
+        SpannableString spannable = new SpannableString(hyperlinkTextView.getText());
+        ClickableSpan clickableSpan = new ClickableSpan() {
+            @Override
+            public void onClick(View widget) {
+                // Start the SecondActivity when the link is clicked
+                Intent intent = new Intent(LoginActivity.this, RegisterActivity.class);
+                startActivity(intent);
```
mccodav1, line 65: I'm not 100% whether finish() should be called after we start new activities; something to keep in mind for the future:

https://stackoverflow.com/questions/10847526/what-is-activity-finish-method-doing-exactly

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
-                e.printStackTrace();
-                Toast.makeText(LoginActivity.this, getString(R.string.internal_error), Toast.LENGTH_LONG).show();
+        /**
+         * Function that runs when login button is pressed
+         * Makes a request to the API and gets a response in return
+         */
+        buttonLoginSubmit.setOnClickListener(view -> {
+            String email = usernameField.getText().toString();
+            String password = passwordField.getText().toString();
+            if (!validateEmail(LoginActivity.this, email)){
```
mccodav1, line 81: Might be useful to also call validatePassword(), so we can enforce some password validity client-side

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
+                return;
             }
+            MyApiService apiService = RetrofitClient.getRetrofitInstance().create(MyApiService.class);
+            Call<ResponseBody> call = apiService.loginUser(new LoginRequest(email, password));
+            call.enqueue(new Callback<ResponseBody>() {
+                /**
+                 * Function that runs when response is returned by the API
+                 * @param call
```
mccodav1, line 92: May be helpful to elucidate params here, too

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -1,99 +1,127 @@
...
+                        } catch (IOException e) {
+                            throw new RuntimeException(e);
+                        } catch (JSONException err){
+                            Log.e("Json Error", err.toString());
+                        }
+                        // open dashboard activity
+                        Intent dashboardIntent = new Intent(LoginActivity.this, IntroActivity.class);
+                        startActivity(dashboardIntent);
+                    }else {
+                        Toast.makeText(LoginActivity.this, "Login failed.", Toast.LENGTH_SHORT).show();
```
mccodav1: May be helpful to parse error responses and display to user; this 'else' blocks indicates a non-success response back from the API, which should contain additional data:

From PosteBackend `LoginView`:
```
        if serializer.is_valid():
            email = serializer.validated_data.get("email")
            password = serializer.validated_data.get("password")
            user = authenticate(request, email=email, password=password)
            if user:
                token, created = Token.objects.get_or_create(user=user)
                return Response(
                    {"result": {"success": True, "token": token.key}},
                    status=status.HTTP_200_OK,
                )

            else:
                return Response(
                    {"result": {"success": False}}, status=status.HTTP_401_UNAUTHORIZED
                )
        else:
            return Response(
                {"result": {"success": False, "errors": serializer.errors}},
                status=status.HTTP_400_BAD_REQUEST,
            )
```

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -46,6 +46,69 @@ public class User {
      */
     private final Collection<Folder> folders;
 
+    /**
+     * The email used for the user's account
```
mccodav1, line 50: Really appreciate the docstrings

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -132,6 +195,7 @@ public void onResponse(@NonNull Call<ResponseBody> call, @NonNull Response<Respo
                     try {
                         String jsonResponse = response.body().string();
                         Log.d("Response", jsonResponse);
+                        user.folders.clear();
```
mccodav1, line 199: Great eye finding the duplication bug!

It might be useful to make a copy of the user's folders before this; that way, if the API call fails for some reason, we can set the folders back to what they were before. Otherwise, if the user loses connection at this moment, it may clear out their application but not rebuild it

mccodav1 (approved):
This is testing working against dev branch of PosteBackend. I like the additions you made and appreciate the bug you found and fixed proactively here. I made some suggestions, but I'm not viewing them as blockers here.

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: Implemented it

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: That was the naming scheme I was kind of going for, editText  is the widget, Text is the data type and EmailAddress is what it contains. Can be fixxed in the future. Needs changes in too many places to fix right now

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: Added finish() wherever I'm starting a new intent.

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: The current method for validatePassword required 2 string sof password and confirm, so I created a new method for it and has since been implemented

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: The reasoning is Im not exactly sure what is in the call here. Can discuss with you and put it in soon

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: Has since been implemented

app/src/main/java/com/example/poste/models/User.java
fmehdad97: Has been implemented. Unable to have tested this fringe case though.

app/src/main/java/com/example/poste/activities/LoginActivity.java
mccodav1: No worries. If that's what works for you, right on; not trying to pick apart coding style, just wanted to make sure there wasn't a typo :)

EDIT: For future reference, if you're using Android Studio, you can click on a field / var name and Shift+F6 or right click > Refactor > Rename. This will rename vars, methods, classes virtually everywhere they're used all at once; it might not catch them in the manifest files, I'm not 100%. But, it comes in SUPER handy sometimes.

PR #9, opened 2023-10-21
Us66.1 data access

dzemlin:
Refactor of the Folder view

To run a test use the commit:
https://github.com/Appyo-Poste/Poste/commit/2df977d326f679b5f0650c934133dc2f657eb433
After clicking login it will instead take you to the folder view with mock data.

Create new post, Folder Settings, and Edit Post buttons, will crash the app simply because the Activity Pages for those functions have not been finished or refactored.
Open post does not function yet because the app needs web connectivity to attempt to open the browser.

Also I added variables to the static PosteApplication class to track the currently selected folder and currently selected post. These can be used from any method to quickly access, copy, or remove data.

Additionally the debug code for creating mock data (in PosteApplication.java) can serve as a blueprint for building the app side data model from the information the API calls return.


app/src/main/java/com/example/poste/PosteApplication.java
```diff
@@ -13,6 +22,8 @@
 public class PosteApplication extends Application {
     private static Context context;
```
mccodav1: I know you didn't write this line, but Android Studio is saying this poses a memory leak:

`Do not place Android context classes in static fields; this is a memory leak`

Is there a workaround?

app/src/main/java/com/example/poste/PosteApplication.java
```diff
@@ -44,6 +99,38 @@ public static void setCurrentUser(User currentUser) {
         PosteApplication.currentUser = currentUser;
```
mccodav1, line 114: This can probably be removed, as we're using a User singleton elsewhere

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -25,32 +16,16 @@
...
-import com.example.poste.api.poste.models.FolderAccess;
 import com.example.poste.api.poste.models.Post;
-import com.example.poste.api.poste.models.User;
-
-import java.util.HashMap;
-import java.net.URL;
-import java.util.Objects;
 
 /**
  * The FolderViewActivity class adds functionality to the activity_folder_view.xml layout
```
mccodav1: It seems there was a lot of confusion when we picked up the project in understand what activities did what (especially since some of them are named in a way that might imply functionality different from what they actually do) -- should we maybe expand on the class docstrings?

This activity is for viewing posts inside a given folder, correct?

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +50,85 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        int currentFolderId = getIntent().getIntExtra("folderId", -1);
-        try {
-            currentFolder = API.getFolderById(currentFolderId);
-
-            // Remove empty text if posts exist in folder
-            if (currentFolder.getPosts().size() > 0) {
-                emptyText.setVisibility(View.GONE);
+        // Create listeners for the folder buttons
+        newBut.setOnClickListener(view -> {
+            Post newPost = new Post(PosteApplication.getCurrentUser().getId(), "", "", PosteApplication.getCurrentUser().getId());
```
mccodav1: I think we should consider making this an API call, rather than creating a post locally. Otherwise, we put our client-side and backend out of sync. Perhaps a more abstracted solution would be to get the data necessary to send a post creation call to the backend, and then either updating all user data (as we do during dashboard / login) or updating the given folder

This could be accomplished via a method on the singleton User, createPost() or some such, that handles making the API call (see User.updateFoldersAndPosts() on dev branch)

Also, it looks like you may need to pull dev into your branch and reconcile changes

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +50,85 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        Post post = postAdapter.getLocalDataSetItem();
-        Intent intent = null;
-        switch (item.getItemId()) {
-            case R.id.ctx_menu_edit_post:
-                intent = new Intent(FolderViewActivity.this, EditPostActivity.class);
-                intent.putExtra("folderId", currentFolder.getId());
-                intent.putExtra("postId", post.getId());
+        deleteBut.setOnClickListener(view -> {
+            if (PosteApplication.getSelectedPost() != null) {
+                PosteApplication.getSelectedFolder().removePost(PosteApplication.getSelectedPost());
```
mccodav1: I like the method call here to abstract functionality, but I think removePost should also handle an API call (see the comment above on adding a post) for the same reasons.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +50,85 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        else {
+            openBut.setVisibility(View.GONE);
+            editBut.setVisibility(View.GONE);
+            deleteBut.setVisibility(View.GONE);
         }
-        return true;
-    }
 
+        // Fill post view (Recycler View)
+        postRecyclerView.setLayoutManager(new LinearLayoutManager(FolderViewActivity.this));
```
mccodav1, line 118: Nice usage of RecyclerView; was this easy to work with?

app/src/main/java/com/example/poste/api/poste/models/Folder.java
```diff
@@ -92,6 +92,15 @@ public void setPosts(List<Post> posts) {
         this.posts = posts;
     }
 
+    /**
+     * Removes a post from the list of posts in this folder
+     *
+     * @param post the post to remove
+     */
+    public void removePost(Post post) {
+        posts.remove(post);
```
mccodav1, line 101: See comments above re: API call for this functionality to avoid client-side / backend sync issues

app/src/main/java/com/example/poste/api/poste/models/Folder.java
```diff
@@ -119,4 +128,8 @@ public boolean update() {
             return API.updateFolder(this.id, this.name, this.ownerId);
         } catch (Exception e) { return false; }
     }
+
+    public void addPost(Post post) {
+        posts.add(post);
```
mccodav1, line 133: See comments above re: API call for this functionality to avoid client-side / backend sync issues

app/src/main/java/com/example/poste/api/poste/models/User.java
```diff
@@ -7,18 +7,24 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
 public class User {
 
     private int id;
     private String email;
     private String username;
     private String password;
+    private List<Folder> folders;
```
mccodav1: Hate for folks to duplicate effort, but these models have been rebuilt (with additional functionality used by the new LoginActivity and RegisterActivity pages) -- if you pull dev into your branch you should see these in the `com.example.poste.models` package

app/src/main/java/com/example/poste/api/poste/models/User.java
```diff
@@ -67,6 +73,18 @@ public void setPassword(String password) {
         if (!update()) { this.password = oldPassword; }
     }
 
+    /**
+     * Adds a folder to the list of folders this user can access
+     * @param folder the folder to be added to the users list of folders
+     */
+    public void addFolder (Folder folder) {
+        folders.add(folder);
```
mccodav1: See above comments re: API calls for these methods

app/src/main/java/com/example/poste/api/poste/models/User.java
```diff
@@ -67,6 +73,18 @@ public void setPassword(String password) {
...
+    }
+
+    public List<Folder> getFolders() {
+        return folders;
```
mccodav1: Despite above comments, I don't think this requires an API call -- we can assume the data has either been already retrieved, or the user can request a data update (screen pull-down-to-refresh or similar functionality)

mccodav1 (changes requested):
Overall I think great work so far on the refactoring. Keep an eye on Slack and dev; changes have been made that should be pulled in here, including new models (and the usage of a singleton User model that should greatly simplify the efforts of retrieving user folders, and providing an outlet for calling update / creation methods to abstract API calls, etc.

app/src/main/java/com/example/poste/PosteApplication.java
dzemlin: In fact this reference goes completely unused and the 2 methods that it is related to are also never called anywhere, SO i am removing them.

Tested the program and it still works without these.
don't know why they were here to begin with.

app/src/main/java/com/example/poste/PosteApplication.java
dzemlin: This a static reference to the current user. It seems throughout the code this is how methods find the current user  Object and subsequently call the methods on it (like get folders). If the currentUser is a singleton, i don't think that changes how it should be referenced and located. 
Or the User class has also been made static?

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: I will add a better description for this class

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: What I intend to do is once the method for updating the back end is created, All of these areas of my code that add a folder/post, or remove a folder/post, will then call the method that updates the API. 
Is that method present yet? if so, where is it? I will hock it in as soon as it is available.
(this response is for all the API related comments above, instead of responding to each individually. )

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: Take a look at the dev branch to get an idea of the implementation. User is a singleton, which means we can access it from anywhere, and always get the same User instance while ensuring only one User is ever created at runtime.

You can get the User instance via User.getUser(), and then call any methods on it as necessary.

The code was previously written in a different fashion; I thought that was part of what we were aiming to refactor, however. The 'new' style is what is currently being used in the new Login and Register activities

dzemlin: Looking over the most resent dev.
Do we now use the static method: User.getUser(); anywhere that the code used previously used PosteApplication.currentUser? 

the old code using this static variable heavily so its should be a new User Story in itself to refactor this out of the code.

app/src/main/java/com/example/poste/api/poste/models/User.java
mccodav1: We can build out the method, and just change its implementation as necessary.

For instance, you might call user.addFolder(folder); then, we could stub that out to either access the data locally (if testing, I suppose) or make the API call. The implementation would then be abstracted from the function call. Really, to avoid duplicate efforts, we should just build out the endpoint and the functionality simultaneously -- that was my understanding of this sprint's goals.

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: That is the intention, I believe, but we are still refactoring.

mccodav1: Also, by using that, we could set a currentFolder and currentPost on user (if you want to use that approach) rather than PosteApplication. That is a design choice however

dzemlin: Im converting mine now to use this new method. but the old "currentUser" needs to get extracted out and the old data model java file need to be refactored out of the code. I will volunteer for this user story next sprint.

dzemlin: Will remove in next user story. Only waiting in order to avoid merge conflicts (in the case that someone used this version in their refactor).

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Discussed below

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Discussed below.

app/src/main/java/com/example/poste/api/poste/models/Folder.java
dzemlin: Discussed below.

app/src/main/java/com/example/poste/api/poste/models/Folder.java
dzemlin: Discussed below.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: Looks like my branch fell behind. I believe I have correct this now and gotten mine back in sync with current development.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: I have made the Client side API calls, Please review and let me know if I am missing something.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: Agreed

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: It looks like you still use these methods in your branch though

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +52,84 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        int currentFolderId = getIntent().getIntExtra("folderId", -1);
-        try {
-            currentFolder = API.getFolderById(currentFolderId);
-
-            // Remove empty text if posts exist in folder
-            if (currentFolder.getPosts().size() > 0) {
-                emptyText.setVisibility(View.GONE);
+        // Create listeners for the folder buttons
+        newBut.setOnClickListener(view -> {
+            Post newPost = PosteApplication.getSelectedFolder().createNewPost(PosteApplication.getSelectedFolder());
```
mccodav1: I think this is a good start. A few considerations:

- When a user clicks the button, this is going to send a request to create a new Post to the backend. However, this simply creates a post with the title 'Untitled', no URL, no description, etc. However...
- The API request looks good! You properly define an endpoint in MyApiService, and create a `CreatePost` request that sends a title, description, url, and folder. As a result...
- Clicking the 'new post' button should start an activity (a client-side page) for creating a post. This would retrieve the additional details: title, description, url. Then, when this is saved, the API call can be made with *all* the details (instead of creating an empty post with an 'Untitled' title)

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +52,84 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        settingsBut.setOnClickListener(view -> {
+            Intent intent = new Intent(FolderViewActivity.this, EditFolderActivity.class);
+            startActivity(intent);
+        });
+
+        // Create listeners for the post buttons
+        openBut.setOnClickListener(view -> {
+            if (PosteApplication.getSelectedPost() != null) {
+                try {
+                    Intent browserIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(PosteApplication.getSelectedPost().getUrl()));
```
mccodav1, line 80: This appears to open the Post. Nice!

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +52,84 @@ protected void onCreate(Bundle savedInstanceState) {
...
-            );
-            postRecyclerView.setAdapter(postAdapter);
-            registerForContextMenu(postRecyclerView);
-        } catch (APIException e) {
-            throw new RuntimeException(e);
-        }
-    }
+        editBut.setOnClickListener(view -> {
+            if (PosteApplication.getSelectedPost() != null) {
+                Intent intent = new Intent(FolderViewActivity.this, EditPostActivity.class);
```
mccodav1, line 91: Does this button edit a post or edit a folder? It seems to call 'EditPost' but we're in Folder view, no? If the intent is to edit a post, I would think this should be contained in a PostView activity -- otherwise, which post is this offering to edit for us?

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,77 +52,84 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    @Override
-    public boolean onContextItemSelected(MenuItem item) {
-        Post post = postAdapter.getLocalDataSetItem();
-        Intent intent = null;
-        switch (item.getItemId()) {
-            case R.id.ctx_menu_edit_post:
-                intent = new Intent(FolderViewActivity.this, EditPostActivity.class);
-                intent.putExtra("folderId", currentFolder.getId());
-                intent.putExtra("postId", post.getId());
+        deleteBut.setOnClickListener(view -> {
```
mccodav1, line 96: Does this button delete a post or delete a folder? It seems to call a method to delete a post, but we're in Folder view, no? If the intent is to delete a post, I would think this should be contained in a PostView activity -- otherwise, which post is this offering to delete for us?

app/src/main/java/com/example/poste/api/poste/models/User.java
```diff
@@ -7,18 +7,24 @@
...
     private String password;
+    private List<Folder> folders;
 
     public User(int id, String email, String username, String password) {
         this.id = id;
         this.email = email;
         this.username = username;
         this.password = password;
+
+        folders = new ArrayList<>();
```
mccodav1: This doesn't hurt anything, but folders is instantiated during the User constructor

app/src/main/java/com/example/poste/api/poste/models/User.java
mccodav1: I think what we want here is for User.addFolder to make the API calls necessary to add a Folder. The key isn't adding a folder to the User's ArrayList, its to populate the backend, so the API call is necessary. If we simply add a folder to the user's arraylist, its only contained in memory, not on the backend. It would be more accurate to make a call to the API here to create the Folder and then update the User's folder list from the backend (or, add it to the backend, and simultaneously add it to the Folder's list locally... but this will be overwritten on data update)

app/src/main/java/com/example/poste/http/CreatePost.java
```diff
@@ -0,0 +1,17 @@
+package com.example.poste.http;
```
mccodav1, line 1: Looks great!

app/src/main/java/com/example/poste/http/MyApiService.java
```diff
@@ -12,6 +13,12 @@ public interface MyApiService {
     @POST("users/")
     Call<ResponseBody> registerUser(@Body RegisterRequest registerRequest);
 
+    @POST("posts/")
+    Call<ResponseBody> createPost(@Body CreatePost createPost);
```
mccodav1: this should also pass the authorization header the way getData does; this is how we tell the backend what user we're talking about (and ensures the user is authorized to perform the action)

In other words, this should pass both a @Body CreatePost, and a @Header("Authorization") String authToken. You can see in the getData method how that authToken is populated

app/src/main/java/com/example/poste/http/MyApiService.java
```diff
@@ -12,6 +13,12 @@ public interface MyApiService {
...
+
+    @DELETE("posts/")
+    Call<ResponseBody> deletePost(@Body DeletePost deletePost);
```
mccodav1: We will also need the auth header here

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -60,6 +69,34 @@ public String getUserPermission() {
         return userPermission;
     }
 
+    /**
+     * Adds a new post the given folder
+     *
+     * @param folder folder to add the new post to
+     */
+    public Post createNewPost(Folder folder) {
```
mccodav1: Similar to what was discussed above:

This method should probably only be called once we have all the appropriate fields: Title, Description, URL, and the request should contain all the fields necessary. You make the API call in 82 with newPost.getDescription(), newPost.getUrl(), but that has just been built without these fields -- they'll never be set. We need to retrieve these from a client-side page (activity)

Take a look at how Login or Register work; an activity is made with fields for the user to fill in, and once they're filled in, the data is retrieved and validated, then sent to the API

This is a great start, however. I would also encourage you to handle failures in some way; otherwise, if the API call fails, nothing will happen, and the user won't know why.

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -60,6 +69,34 @@ public String getUserPermission() {
...
+        call.enqueue(new Callback<ResponseBody>() {
+            @Override
+            public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                posts.add(newPost);
+            }
+            @Override
+            public void onFailure(Call<ResponseBody> call, Throwable t) {
+            }
+        });
+        if (posts.contains(newPost)) {
```
mccodav1: This can potentially throw a NullPointerException elsewhere; it is called in FolderViewActivity (and perhaps elsewhere?) and assigned to a variable, which is then set as SelectedPost, which is used elsewhere, etc.

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -67,6 +76,24 @@ public String getUrl() {
      */
     public String getId() { return id; }
 
+    /**
+     * Deletes this post from the data model and backend
+     */
+    public void deletePost(Folder parentFolder) {
```
mccodav1: I would contend this should be a method on Folder, rather than on Post. A Post is deleted from a folder, and a folder manages its children (the posts within)

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -67,6 +76,24 @@ public String getUrl() {
...
+        Post delPost = this;
+        MyApiService apiService = RetrofitClient.getRetrofitInstance().create(MyApiService.class);
+        Call<ResponseBody> call = apiService.deletePost(new DeletePost(delPost.getId()));
```
mccodav1: Good start; remember we'll also need to pass the auth token in the header (see getData for more context)

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -67,6 +76,24 @@ public String getUrl() {
...
+        call.enqueue(new Callback<ResponseBody>() {
+            @Override
+            public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                parentFolder.getPosts().remove(delPost);
```
mccodav1: Remember that a response can indicate failure (such as 404, 500, etc) -- would be useful to handle these error responses gracefully

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -67,6 +76,24 @@ public String getUrl() {
...
+            }
+            @Override
+            public void onFailure(Call<ResponseBody> call, Throwable t) {
```
mccodav1: As above, but this indicates a failure to send the API request at all (connectivity, etc)

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -72,7 +72,7 @@ public static User getUser() {
      * Adds a folder to the user's folders.
      * @param newFolder the folder to add to the user's folders.
      */
-    private void addFolder(Folder newFolder) {
+    public void addFolder(Folder newFolder) {
```
mccodav1, line 140: Should ideally be an API call, as above

mccodav1 (changes requested):

app/src/main/java/com/example/poste/PosteApplication.java
dzemlin: These are the legacy methods unrelated to this branch. I will remove them on a different User Story. 
they are still used in other students branches (login and editPost for example). I did not add these methods. Removing them before the other branches are pulled will likely cause merge conflicts I just changed the import to be explicit so that they would not be accidentally mistaken for the new User access methods. Removing these will be done in https://tree.taiga.io/project/dzemlin-poste/task/145

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: This is the old User.java file, that i edited before pulling dev. it is no longer needed and this whole java file will be deleted in my next User Story.

we currently have 2 User.java files in 2 different places:
example/poste/api/poste/models/User.java (the old one)
example/poste/models/User.java                  (the new one)

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Agreed, Breaking up the API call is unnecessary, I Missunderstood and though you wanted me to split them up. 
In this case, since EditPost and create know poste are done from the same Activity Page then the API call should not be made here.
I will change it simple to open the EditPostActivity with no selected post. I will touch based with the student who refactored EditPostActivity to make sure we are passing the right data for the activity to know if it is pushing changes to the API or making a new Post in it.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: FolderViewActivity is not the view of ALL your folders (that is the DashboardActivity).
FolderViewActivity is the view of what is inside 1 particular folder (i.e. its list of posts).
So this button will open EditPostActivity with the post selected from this screen.
There is no PostViewActivity. FolderView is effectivly the bottom layer of the hierarchy, unless you consider EditPost & EditFolder to be a layer

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: See explanation above about the absence of a "PostView".

It would delete whichever post you have selected.
You tap a post from the recycler view, then tap delete. 
Then that post will know who's boss.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: Same issue as before, this is the old User.java file. it is soon to be incinerated.

dzemlin: Same issue as before, this is the old User.java file. it is soon to be incinerated.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: Same issue as before, this is the old User.java file. it is soon to be incinerated.

app/src/main/java/com/example/poste/http/MyApiService.java
dzemlin: I'm gonna need some one on one time with you tomorrow on this.
I don't think I understand what to do here. reach out on slack and we can set a time.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: I was a little unsure on this. because app-side. Folders contain posts and a post would be deleted by simply removing it from the folder's list.
But I saw that on the backend, the post has control. From what I could tell the folders in the data base do not even know what posts they have, but it is the post that knows to which folder it belongs. SO i was trying to mimic that post centric management here. But I'm fine doing it the other way too. I just wanted to explain why I did it like this. 

TLDR: Cool. Moving it now.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: This too I will need a little more explanation on when we meet.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Yes. In due time. From what we discussed though I'm focusing on the happy day cases first. for the sake of moving toward MVP.

app/src/main/java/com/example/poste/models/User.java
dzemlin: All i did here was make this public instead of private (for the debugger method).

This is the method you made for building the data modal from the API response, NOT one that the program will use to make a new folder.
so once a folder is built from the API response, this is simply telling the app to include it in the list of folders. 
at least that's what I figured it was for.

image

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: Understood. Just want to make sure we are each incrementally moving our own tasks over to the new style so we can eventually remove, and confusion is mitigated.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
mccodav1: Exactly. We could even split up the activities -- even if they appear almost exactly identical -- as it may help abstracting out some concepts. For instance, while they appear the same to the user, creating a new post would have no fields filled out (blank slate) whereas editing a post should probably have the existing fields populated. Also, while creating a post is a straightforward, an edit would need a different API call to ensure we are truly editing the backend post and not creating a duplicate. It sounds like you know where to go with this, so I'll mark this as resolved.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
mccodav1: Roger -- I wasn't familiar with RecyclerView. I had imagined clicking a post would open it, and the options would be a result of a long press or some such. That perhaps is room for future UX improvement (so a user doesn't have to tap twice to open a post -- I seem to recall we had a requirement on number of clicks to get to a post, but don't remember what it is offhand). If what this *actually* does is allow you to highlight a post and then click open / edit / delete, that makes sense. Thank you for the explanation.

app/src/main/java/com/example/poste/api/poste/models/User.java
mccodav1: Roger. Are we still making changes to these old models?

app/src/main/java/com/example/poste/http/MyApiService.java
mccodav1: Feel free to DM me. There's also a new PR up that shows a post deletion call. Happy to discuss or make a video.

app/src/main/java/com/example/poste/models/Post.java
mccodav1: The backend is a slightly different beast. The database requires ForeignKey relations: models relate to others via ForeignKey (unsure if you've taken a database course or not; Django uses ORM, which is Object-Relational-Mapping, so while we don't directly interact with the database, Django does it for us -- and it uses SQL standards, so concepts relate)

A Post has a ForeignKey to a Folder, which is the way of stating a Post belongs to a Folder; in this way, a Post has one Folder, but a Folder can have many posts. Django *does* have a way to go "the other direction" though:

```
folder = Folder.objects.get(id=some_id)
posts_in_folder = folder.post_set.all()
```

In reality though, after coding post deletion, it seemed logical to place the methods on User. Your mileage may vary.

app/src/main/java/com/example/poste/models/Post.java
mccodav1: These will be mostly cut-and-paste, but always happy to chat about it. I think you'll be pleasantly surprised at the ease (and how repetitious it is). One day I may even abstract it all away into a single method call...

app/src/main/java/com/example/poste/models/User.java
mccodav1: I see. No worries. I thought you were using this in another fashion, misunderstood.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Not to bad. still trying to figure out some of the cosmetic features.

app/src/main/java/com/example/poste/api/poste/models/User.java
dzemlin: no, not intentionally.
I will extract them while building the MVP so they don't get accidently used again

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
mccodav1: I see. So we're expecting a user to click a post, which only selects it, and then click the edit button? Understood. I did not know that was the intended functionality.

app/src/main/java/com/example/poste/models/Post.java
mccodav1: Roger. I would ask that you at least introduce a simple `if (response.isSuccessful()) { ... } else { ... }` block. While I understand building towards 'happy day' cases, we want to ensure we aren't needlessly introducing hard-to-troubleshoot bugs (where it appears things were successful, with no logging or display otherwise, even if they weren't)

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -97,6 +98,25 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+     *
+     * @param post the post you want to delete
+     */
+    public void deletePost(Post post) {
+        MyApiService apiService = RetrofitClient.getRetrofitInstance().create(MyApiService.class);
+        Call<ResponseBody> call = apiService.deletePost(new DeletePost(post.getId()));
+        call.enqueue(new Callback<ResponseBody>() {
+            @Override
+            public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                posts.remove(post);
```
mccodav1: Please add a simple `if (response.isSuccessful()) { ... } else { ... } block`. While I understand building towards 'happy day' cases, we want to ensure we aren't needlessly introducing hard-to-troubleshoot bugs (where it appears things were successful, with no logging or display otherwise, even if they were

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -97,6 +98,25 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+            }
+            @Override
+            public void onFailure(Call<ResponseBody> call, Throwable t) {
+            }
```
mccodav1: Please also introduce logging, toast messages, etc... something to help troubleshoot when this function is called

mccodav1 (commented):
Couple small things, but I think this is coming together nicely. We can chat today about the other stuff, whenever you're free

app/src/main/java/com/example/poste/PosteApplication.java
```diff
@@ -23,15 +33,75 @@ public class PosteApplication extends Application {
...
+        }
     }
 
     /**
      * Get the currently logged-in user
      *
      * @return The User object representing the currently logged-in user
      */
-    public static User getCurrentUser() {
+    public static com.example.poste.api.poste.models.User getCurrentUser() {
```
mccodav1, line 104: Thanks for changing the model. Ideally, everywhere that calls `getCurrentUser()` should be calling `User.getUser()` (the singleton getter)... until that time, can you have this method return `User.getUser()`? This will, at the very least, ensure we don't somehow use / access / do something with an incorrect User instance.

app/src/main/java/com/example/poste/PosteApplication.java
```diff
@@ -40,25 +110,39 @@ public static User getCurrentUser() {
      *
      * @param currentUser The User object representing the currently logged-in user
      */
-    public static void setCurrentUser(User currentUser) {
+    public static void setCurrentUser(com.example.poste.api.poste.models.User currentUser) {
```
mccodav1, line 113: As a result of above, this can be removed, and anywhere in the code that uses it can be deleted outright (as the user will never need to be set)

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -302,4 +304,62 @@ private String parseError(Response<ResponseBody> response) {
...
+            @Override
+            public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                if (response.isSuccessful()) {
+                    updateFoldersAndPosts(context);
+                }
+                else {
+                    try {
+                        Toast.makeText(context, response.errorBody().string(), Toast.LENGTH_LONG);
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
```
mccodav1: (Copy pasted from above -- sorry)


Careful with these -- it'll crash the app. You could use a generic error message, or try to parse it out -- mind how the response is given from the backend. Or, consider using parseError -- a method also in User. It is setup to parse error messages from another API call. Assuming the responses are similar, it may work for this, too. The idea is they should all respond the same way, but I can't guarantee all implementations are the same (yet...)

At the very least, if trying to parse the error message fails, I'd suggest making a generic toast (and please log the response) rather than crashing

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -302,4 +304,62 @@ private String parseError(Response<ResponseBody> response) {
...
+            }
+        });
+    }
+
+    /**
+     * Deletes a given post from a given folder in the data model and backend
+     *
+     * @param post the post you want to delete
+     */
+    public void deletePost(Context context, Post post, Folder folder) {
```
mccodav1, line 339: I thought I mentioned; if not, I apologize. The PR I put up a few days ago has a DeletePost implementation already, with the callback we discussed in the huddle this morning. Your choice how you want to press forward. If you want to steal that implementation, feel free to take a gander at it.

mccodav1 (changes requested):

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -302,4 +304,62 @@ private String parseError(Response<ResponseBody> response) {
...
+            @Override
+            public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                if (response.isSuccessful()) {
+                    updateFoldersAndPosts(context);
+                }
+                else {
+                    try {
+                        Toast.makeText(context, response.errorBody().string(), Toast.LENGTH_LONG);
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
```
mccodav1: Careful with these -- it'll crash the app. You could use a generic error message, or try to parse it out -- mind how the response is given from the backend. Or, consider using parseError -- a method also in User. It is setup to parse error messages from another API call. Assuming the responses are similar, it may work for this, too. The idea is they should all respond the same way, but I can't guarantee all implementations are the same (yet...)

At the very least, if trying to parse the error message fails, I'd suggest making a generic toast (and please log the response) rather than crashing

app/src/main/java/com/example/poste/http/MyApiService.java
dzemlin: Added header

app/src/main/java/com/example/poste/models/Folder.java
dzemlin: Rewritten as directed in discussion this morning

app/src/main/java/com/example/poste/models/Folder.java
dzemlin: Method now returns void. no longer an issue

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Added base "if (responce.isSuccessful()) {} else {} " branches and Toast messages to provide basic error handling,

app/src/main/java/com/example/poste/models/Folder.java
dzemlin: Added base "if (responce.isSuccessful()) {} else {} " branches and Toast messages to provide basic error handling,

app/src/main/java/com/example/poste/models/Folder.java
dzemlin: Added Toast messages

app/src/main/java/com/example/poste/PosteApplication.java
dzemlin: I will extract this in the sprint 3 so it does trick me when building the MVP. 
If I did it this sprint it will likely bring up merge conflicts.
I already have a US task for this I am assigned to -> https://tree.taiga.io/project/dzemlin-poste/task/145

app/src/main/java/com/example/poste/PosteApplication.java
dzemlin: I will extract this in the sprint 3 so it does trick me when building the MVP.
If I removed it this sprint it will likely bring up merge conflicts.
I already have a US task for this I am assigned to -> https://tree.taiga.io/project/dzemlin-poste/task/145

app/src/main/java/com/example/poste/models/User.java
dzemlin: Switched to generic error toast.

app/src/main/java/com/example/poste/models/User.java
dzemlin: Switched to generic error toast.

app/src/main/java/com/example/poste/models/User.java
dzemlin: once we get all the PR's together into dev I will mimic your callback work and apply it to all API calls for the MVP.

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: As above:

Please be specific in user stories / tasks so that individual needs aren't forgotten. For instance, specifying *which* code to remove, refactor, etc.

I'd much rather see this here so we aren't changing the same code twice, and since this code is already changing here -- also, merge conflicts are the result of multiple developers working on the same code, so no merge conflict looks to arise from this. 

I'll mark this as resolved for now, but have updated the Taiga task. Please consider for the future.

app/src/main/java/com/example/poste/PosteApplication.java
mccodav1: As above:

Please be specific in user stories / tasks so that individual needs aren't forgotten. For instance, specifying *which* code to remove, refactor, etc.

I'd much rather see this here so we aren't changing the same code twice, and since this code is already changing here -- also, merge conflicts are the result of multiple developers working on the same code, so no merge conflict looks to arise from this.

I'll mark this as resolved for now, but have updated the Taiga task. Please consider for the future.

app/src/main/java/com/example/poste/models/User.java
mccodav1: It might be more efficient if we merge yours in first and I resolve the conflicts on my branch, since it will take some work integrating the callbacks, etc. The concern is we won't be able to get them all together into dev in the first place -- there will be some large merge conflicts.

mccodav1 (approved):

mccodav1:
Looks like there's a merge conflict with ` app/src/main/java/com/example/poste/api/poste/models/User.java` that just pertains to imports. They can all safely be included.

PR #12, opened 2023-10-24
Implement deletePost

mccodav1:
This commit implements post deletion from the client.

- a `deletePost` method was added to `MyApiService` which defines the route DELETE:"posts/" to make an API call to delete the Post from the backend

- `deletePostFromServer` method added to `com.example.poste.models.User` class, which makes the API call to the server to delete a Post. It takes a Post model (defined in `com.example.poste.models.Post) and a callback as parameters. Callback will be discussed below.

- `PostDeletionCallback` was added to `com.example.poste.calbacks`. Please review the code for usage details; I've added extensive docstrings. Since API calls are made asynchronously, we cannot depend on return values like we might in a standard Java application. A 'callback' informs the calling code of asynchronous success or failure -- this way, we can decide what to do if a request succeeds or fails without waiting on the response.

- `DebugUtils` has had some extra utilities added

- RetrofitClient was slightly modified to add a logger

TL;DR: To add post deletion to an activity:
1) Define a PostDeletionCallback in your activity, implementing `onSuccess` and `onError` methods. 2) Call `User.getUser().deletePostFromServer`, passing a Post model and the callback. 3) If the call succeeds, the `onSuccess` method will run. Otherwise, the `onError` method will run. 4) If you have questions, look at the `User.deletePostFromServer` docstring


app/src/main/java/com/example/poste/http/MyApiService.java
```diff
@@ -28,4 +29,32 @@ public interface MyApiService {
...
+     *     <li><a href="https://square.github.io/retrofit/2.x/retrofit/index.html?retrofit2/http/Header.html">Retrofit Headers</a></li>
+     * </ul>
+     *
+     * @param authToken Authorization token in the format "Token &lt;token&gt;"
+     * @param id ID of the post to be deleted in the format "id: &lt;id&gt;"
+     */
+    @DELETE("posts/")
+    Call<ResponseBody> deletePost(
+            @Header("Authorization") String authToken,
+            @Header("id") String id
```
mccodav1: I don't know why I did it this way. After thinking on it, it makes more sense to use id in the URL path rather than in a header or body. I have a branch to fix this, pending update tomorrow, as I need to also update the backend.

app/src/main/java/com/example/poste/http/RetrofitClient.java
```diff
@@ -10,7 +11,11 @@ public class RetrofitClient {
 
     public static Retrofit getRetrofitInstance() {
         if (retrofit == null) {
-            OkHttpClient client = new OkHttpClient.Builder().build();
+            HttpLoggingInterceptor logging = new HttpLoggingInterceptor();
+            logging.setLevel(HttpLoggingInterceptor.Level.HEADERS);
+            OkHttpClient client = new OkHttpClient.Builder()
+                    .addInterceptor(logging)
```
mccodav1, line 17: This adds logging to our HTTP request client, for debugging purposes.

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -246,7 +247,100 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+     *
+     * User user = User.getUser();
+     * user.deletePost(postToBeDeleted, callback);
+     * }
+     * </pre>
+     *
+     * @param post The Post model to delete.
+     * @param callback The callback to be invoked after attempting to delete the post.
+     */
+    public void deletePostFromServer(Post post, PostDeletionCallback callback) {
```
mccodav1, line 338: In consideration, we should probably use callbacks in more places.

I found this after googling why I wasn't having success returning booleans from functions, and it has to do with the asynchronous calls happening inside.

A callback effectively allows you to specify what you'll do if an async event (such as an API call) succeeds or fails (really, you can define any number of situations). When you make the request, you can make certain callbacks to effect change.

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -246,7 +247,100 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+            callback.onError("Post ID is null");
+            Log.d("UserDebug", "Error deleting post: post ID is null");
+            return;
+        }
+        
+        MyApiService apiService = RetrofitClient.getRetrofitInstance().create(MyApiService.class);
+        Log.d("UserDebug", "Sending delete request to API for Post ID " + id);
+        Call<ResponseBody> call = apiService.deletePost(getTokenHeader(), id);
+
+        call.enqueue(new Callback<ResponseBody>() {
```
mccodav1, line 351: Here, we see we've actually been using Callbacks all along, without much consideration to their meaning

app/src/main/java/com/example/poste/http/MyApiService.java
atreichl: I agree with the ID being in the URL path.

app/src/main/java/com/example/poste/http/RetrofitClient.java
atreichl: This is a good idea.

atreichl (changes requested):
It looks good to me, I'll approve it once you change the id to part of the URL path.

mccodav1:
> It looks good to me, I'll approve it once you change the id to part of the URL path.

Done.

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -109,7 +109,6 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
                             Log.e("Json Error", err.toString());
                         }
                         // open dashboard activity
-                        User.getUser().updateFoldersAndPosts(LoginActivity.this);
```
atreichl: This is probably caused by newer merges happening since you last merged dev in, but if this line is removed dashboard will not have any folders when it launches.

atreichl (changes requested):
fix the one minor error I mentioned and it's good to go.

app/src/main/java/com/example/poste/activities/LoginActivity.java
mccodav1: I guess I assumed this was occurring in DashboardActivity after a discussion I had with Fayez, but I'll add it back here.

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -49,47 +51,57 @@ protected void onCreate(Bundle savedInstanceState) {
 
         DebugUtils.logUserFoldersAndPosts(User.getUser());
 
-        EditText linkToPost = findViewById(R.id.editTextLinkToPost);
+        EditText postLink = findViewById(R.id.editTextPostLink);
```
mccodav1, line 54: Just to keep naming conventions alike

app/src/main/java/com/example/poste/http/DeletePost.java
mccodav1, line 1: Similar to EditPost or DeleteFolder, we'll want to pass in just a String here (for path ID) rather than creating an object (nothing goes in body)

app/src/main/java/com/example/poste/http/utils.java
mccodav1, line 1: Extracted this from User into its own static method

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -82,12 +81,11 @@ public String getUserPermission() {
     }
 
     /**
-     * Returns a copy of the list of posts in the folder. This is a copy, so modifying the returned
-     * list will not modify the folder's posts.
-     * @return a copy of the list of posts in the folder.
+     * Returns a List of the posts in the folder.
```
mccodav1, line 84: I like the idea of immutable return values here (returning a copy, instead of the actual list) -- however, I think this was causing some confusion and we were attempting to modify the returned lists of folders / posts. Until we can get on the same page and refactor, this will solve some issues.

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -177,17 +176,17 @@ public Post getPost(String postId) {
...
     /**
-     * Returns a copy of the list of the user's folders. This is a copy, so modifying the returned
-     * list will not modify the user's folders.
-     * @return a copy of the list of the user's folders.
+     * Returns List of the user's folders.
+     * @return a List of the user's folders.
      */
     public List<Folder> getFolders() {
-        return new ArrayList<>(folders);
+        return folders;
```
mccodav1, line 183: I like the idea of immutable return values here (returning a copy, instead of the actual list) -- however, I think this was causing some confusion and we were attempting to modify the returned lists of folders / posts. Until we can get on the same page and refactor, this will solve some issues.

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -202,7 +201,7 @@ public String getToken() {
      * "Authorization": "Token A12345"
      * @return the user's token, prepended with "Token "
      */
-    public String getTokenHeader() {
+    public String getTokenHeaderString() {
```
mccodav1, line 204: Refactored name for clarity

atreichl (approved):
Everything looks good to me now.

build.gradle
```diff
@@ -1,6 +1,6 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id 'com.android.application' version '8.1.1' apply false
-    id 'com.android.library' version '8.1.1' apply false
+    id 'com.android.application' version '8.0.2' apply false
+    id 'com.android.library' version '8.0.2' apply false
     id 'org.jetbrains.kotlin.android' version '1.8.10' apply false
```
dzemlin: shouldn't we leave the build version at 8.1.1 since it forces us to update to that every time we run?

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -106,46 +131,20 @@ public void setLastName(String lastName) {
...
-    }
-
 
     /**
      * Adds a folder to the user's folders.
      * @param newFolder the folder to add to the user's folders.
      */
-    public void addFolder(Folder newFolder) {
+    private void addFolder(Folder newFolder) {
         folders.add(newFolder);
```
dzemlin, line 140: if this causse merge conflicts with my debug data model method in PosteApplicatin.java, feel free to comment out that method.

app/src/main/java/com/example/poste/models/User.java
dzemlin: Fair enough

dzemlin (approved):
Looks good to me. but I think a third set of eyes would be a good idea too.

build.gradle
mccodav1: I have the opposite problem and I imagine might be something local to your install; the project has historically been on 8.0.2. I can change it to 8.1.1 but that will just cause the reverse issues for others...

app/src/main/java/com/example/poste/models/User.java
mccodav1: I think we might be using the term 'merge conflict' differently; can you elaborate what you mean?

PR #13, opened 2023-10-25
Us64T111 Implement editPost functionality

fmehdad97:
The way that the activity knows which post it is editing is by passing in the postID through .putExtra in the intent.

Maybe some of the editing functionality might be better to be refactored to the User model to make the activity leaner? I am open to suggestions and discussion from reviewer.


app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -2,52 +2,32 @@
...
-import com.example.poste.api.poste.models.FolderAccess;
-import com.example.poste.api.poste.models.Post;
-
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.stream.Collectors;
+import com.example.poste.http.EditPostRequest;
+import com.example.poste.http.LoginRequest;
```
mccodav1: Unused here I think

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -2,52 +2,32 @@
...
+
+import okhttp3.ResponseBody;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
 
 /**
  * The EditPostActivity class adds functionality to the activity_edit_post.xml layout
+ *
+ * The activity that calls this activity, do a .putExtra to pass the postId
```
mccodav1, line 29: Can you add an example here, showing how to call the activity with the putExtra? Do we need to send just postId, or also folderId?

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
-                WindowManager.LayoutParams.FLAG_FULLSCREEN);
-        getSupportActionBar().hide();
+        DebugUtils.logUserFoldersAndPosts(User.getUser());
 
-        // Set the activity layout
-        setContentView(R.layout.activity_edit_post);
+        EditText linkToPost = findViewById(R.id.editTextLinkToPost);
+        EditText postTitle = findViewById(R.id.editTextPostTitle);
+        EditText postDescription = findViewById(R.id.editTextPostDescription);
+
```
mccodav1, line 55: I noticed that we don't have editTextPostTags hooked up to anything; should we? What's the desired functionality look like? Should we add an endpoint and include this in API calls? Deferring to you for your take on this

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
-            currentFolderTitle = String.format("(%d) %s", currentFolder.getId(), currentFolder.getName());
-            userFolders = API.getFoldersForUserId(PosteApplication.getCurrentUser().getId());
-            for (Folder folder: userFolders.keySet()) {
-                folderIdNameMap.put(String.format("(%d) %s", folder.getId(), folder.getName()), folder.getId());
+        Folder currentFolder = User.getUser().getFolder(folderId);
+        if (currentFolder == null){
+            currentPost = null;
+            Log.e("Error", "onCreate: Folder not found");
+        } else {
+            currentPost = User.getUser().getFolder(folderId).getPostFromFolder(postId);
```
mccodav1: I like the chaining functionality here, but I wonder if a User method `getPost(String postId)` or `getPost(String folderId, String postId)` might be a simpler implementation? 

Since Posts are unique by ID, we don't *really* need to know the folder at this point; we could simply search for it through all folders. This is more performant, but the difference should be very small; up to you. But, abstracting the method call away might be nice. It would also give a way to handle a null folder OR post (because it can just return null if either post OR folder is null)

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
-        try {
-            currentPost = API.getPostById(getIntent().getIntExtra("postId", 0));
-            currentFolder = API.getFolderById(getIntent().getIntExtra("folderId", 0));
-            currentFolderTitle = String.format("(%d) %s", currentFolder.getId(), currentFolder.getName());
-            userFolders = API.getFoldersForUserId(PosteApplication.getCurrentUser().getId());
-            for (Folder folder: userFolders.keySet()) {
-                folderIdNameMap.put(String.format("(%d) %s", folder.getId(), folder.getName()), folder.getId());
+        Folder currentFolder = User.getUser().getFolder(folderId);
+        if (currentFolder == null){
+            currentPost = null;
```
mccodav1, line 65: I like the null check here! You might consider also checking if getFolder returns null; otherwise, a NullPointerException will crash the app here

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
-        spinner.setSelection(adapter.getPosition(currentFolderTitle));
-
-        // Cancel button click handler
-        cancelBtn.setOnClickListener(view -> {
-            // Send back to folder view
-            sendToDashboard();
+        buttonCancelChanges.setOnClickListener(view -> {
+            Intent returnToFolderIntent = new Intent(EditPostActivity.this, EditPostActivity.class);
+            finish();
+            returnToFolderIntent.putExtra("postID",postId);
```
mccodav1, line 81: Ah, I was hoping for an example, thank you. This might be beneficial in the docstring too

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
-                if (currentFolder.getId() != selectedFolderId) {
-                    API.removePostFromFolder(currentPost.getId(), currentFolder.getId());
-                    API.addPostToFolder(currentPost.getId(), selectedFolderId);
-                }
-
-                // Send back to folder view
-                sendToDashboard();
-            } catch (APIException e) {
-                throw new RuntimeException(e);
+        buttonSaveChanges.setOnClickListener(view -> {
```
mccodav1, line 85: is there a way to exit the activity?

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
+                call.enqueue(new Callback<ResponseBody>() {
+                    @Override
+                    public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
+                        if (response.isSuccessful()){
+                            currentPost.setTitle(title);
+                            currentPost.setDescription(description);
+                            currentPost.setUrl(url);
+                            Toast.makeText(EditPostActivity.this,"Edit post successful!", Toast.LENGTH_SHORT).show();
+                        } else {
+                            Toast.makeText(EditPostActivity.this,"Post to edit not found!", Toast.LENGTH_SHORT).show();
```
mccodav1: Might be beneficial to revisit this and pull the error that the server is returning to the client; while 'not found' is one possible cause, there could be others (token expired, not authorized or no access, etc)

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -59,84 +39,73 @@ public class EditPostActivity extends AppCompatActivity {
...
+                        }
+                    }
+
+                    @Override
+                    public void onFailure(Call<ResponseBody> call, Throwable t) {
+                        Toast.makeText(EditPostActivity.this, "Edit failed due to server error", Toast.LENGTH_SHORT).show();
```
mccodav1: I was reading on this yesterday and the onFailure here is called when the Response itself fails; could be due to connectivity or client-side failure, too, but generally means the connection to the server never happened

app/src/main/java/com/example/poste/activities/LoginActivity.java
```diff
@@ -108,9 +108,13 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
                         } catch (JSONException err){
                             Log.e("Json Error", err.toString());
                         }
-                        // open dashboard activity
+                        // open dashboard activity (CURRENTLY TAKES BACK TO INTRO, TAKING TO DASHBOARD CRASHES)
                         Intent dashboardIntent = new Intent(LoginActivity.this, IntroActivity.class);
                         finish();
+                        // The updateFoldersandPost should be called in Dashboard
```
mccodav1: I think I might recommend uncommenting these lines and, if necessary, allowing the app to crash (preferably with logging) so we can determine why it crashes and how to resolve. Otherwise this sort of obfuscates the issue and control flow.

Not expecting you to debug / solve this in THIS pull request, but I'm weary of introducing unrelated control flow changes

app/src/main/java/com/example/poste/http/EditPostRequest.java
```diff
@@ -0,0 +1,13 @@
+package com.example.poste.http;
+
+public class EditPostRequest {
```
mccodav1, line 3: Are we doing anything with tags?

app/src/main/java/com/example/poste/http/MyApiService.java
```diff
@@ -15,6 +16,11 @@ public interface MyApiService {
     @POST("login/")
     Call<ResponseBody> loginUser(@Body LoginRequest loginRequest);
 
+    @PATCH("posts/")
```
mccodav1, line 27: Nicely done.

app/src/main/java/com/example/poste/models/Folder.java
```diff
@@ -52,6 +53,16 @@ public String getTitle() {
         return title;
     }
 
+    public Post getPostFromFolder(String postId){
```
mccodav1, line 66: Nicely done. You might consider placing a method on User, perhaps two with one overloaded with FolderID; that's a matter of preference and abstraction level though.

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -139,6 +139,22 @@ private void addFolder(Folder newFolder) {
         folders.add(newFolder);
     }
 
+    /**
+     * Adds a folder to the user's folders.
+     *
+     * @param folderId string containing the id of the folder
+     * @return List of posts in the folder
```
mccodav1, line 148: Perhaps a copy-paste mismatch, this docstring doesn't appear to match the method

mccodav1 (approved):
I have a handful of suggestions / comments, but nothing I consider blocking on this PR. You might consider implementing some of these before merging over; alternatively, perhaps create User Stories in the backlog to resolve individual issues.

In industry it would not be uncommon to turn comments / suggestions here into user stories to be utilized in a future sprint, and might help us with sprint planning if we build them proactively

Well done

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: Has since been optimized, good catch! leftover from copying code and auto-import.

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: I have included the example.

Current implementation needs both folderId and postId.
I agree it is possible to do it with only postId... For some reason it just didn't occur to me when I was coding this

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: So the Tags was there from the prototype. I made the layout following the proto, and then during implementation I found tags wasnt implemented anywhere else... So I left it hanging for now. We can definitely incorporate it now, or after MVP when we begin to implement additional features such as searching for folders by name or posts by title, description or tags. It is based on a brief conversation in chat we had regarding adding searchable tags in the program. The tags could be user input, but my thought was it would probably be an ML classification algo. We can discuss more?

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: I can do the search based on only postID... I'll implement it ASAP.
But Im confused on what you mean by abstracting the method call away. Can you elaborate? I'll shoot a message over Slack.

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: Has since been added, thanks to your review

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: Cancel should take me back to the folder. I wasn't sure how to do that without the app crashing since the folder refactor hadnt been done yet. I'll clean that up and test it and commit again

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: currentFolder == null checks if getFolder returns null, since currentFolder gets the return value assigned to it from getFolder...
Unless I'm missing something?

app/src/main/java/com/example/poste/http/EditPostRequest.java
fmehdad97: Not at the moment. It's a future feature, I will comment//hide it in the layout for now

app/src/main/java/com/example/poste/models/Folder.java
fmehdad97: Partially done, can expand in future

app/src/main/java/com/example/poste/models/User.java
fmehdad97: Yes, fixed it since

app/src/main/java/com/example/poste/activities/LoginActivity.java
fmehdad97: Sounds good. I will uncomment and let it crash then

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: I've made the message more ambigious since to be more accurate

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: Added. 
I did not think of those errors before. So I did not include those errors, but it has since been abstracted to show the error message sent by the server

app/src/main/java/com/example/poste/activities/EditPostActivity.java
mccodav1: Sorry, I think I was unclear.

In line 54, we getStringExtra(); what if this value isn't passed? Is it possible to be null? Or rather, what happens if we pass an invalid value to `getFolder`?

app/src/main/java/com/example/poste/activities/EditPostActivity.java
mccodav1: I see. Just curious. No worries.

app/src/main/java/com/example/poste/activities/EditPostActivity.java
mccodav1: I simply meant calling something like `user.getPostById(<postid>)` and handling all the logic there, or calling later functions there if necessary

app/src/main/java/com/example/poste/activities/EditPostActivity.java
mccodav1: We should be able to parse out the error message from the server response directly and show that (or use it to determine what toast to make, at the least) -- I only see "Post to edit not found" or "Edit failed due to server error". 

Incidentally, the onFailure means a _request_ error, which likely indicates a connectivity error (means no reply was received from the server at all)

app/src/main/java/com/example/poste/activities/LoginActivity.java
mccodav1: Best practice might be to create a user story to resolve, including specifics on this issue and what's causing it. That backlog of bugs will help us in this sprint.

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: Has since been implemented

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: OHH yes that makes sense. I'll make a note of it to get to that ASAP.

PR #16, opened 2023-10-28
Integrate dashboard

mccodav1:
A handful of changes, refactoring, and code-cleanup that integrates Dashboard, FolderView, LoginView, and general flow. 

The big update here is that this adds an UpdateCallback to updateFoldersAndPosts. This is explained in code documentation, but the short version is this: We couldn't trust that updating user data would accomplish in time to perform various tasks (updating folders, posts, etc) -- a callback is a solution to this problem. We define what we want to happen if the update succeeds or fails (for instance, populating the folders to display to the user or displaying an error message) in the callback, and then pass that callback to the update method. This way, we can continue along in our program, and when we get a response back, we handle it accordingly.

This is similar to how the API service already functions, and is a key for handling asynchronous programming.

This has been tested locally against an integration backend branch which I have also opened a MR for; these should be merged at the same time to ensure compatibility. https://github.com/Appyo-Poste/PosteBackend/pull/18

Video demonstrating proper loading of folders on Dashboard, creation, deletion, and our current pain point: creation of a post
https://drive.google.com/file/d/1QDENKLWQ8gyBy5q8xL6ZgM-c2_r_J0im/view?usp=drive_link

### DIFF
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -17,7 +17,6 @@ import android.widget.PopupMenu;
 import android.widget.Spinner;
 import android.widget.Toast;
 
-import androidx.recyclerview.widget.AdapterListUpdateCallback;
 import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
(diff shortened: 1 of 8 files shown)


dzemlin (approved):
Looks great.
I approve.

PR #17, opened 2023-10-29
Us124t163

mccodav1:
Add content sharing to Poste from outside app

This implements the 'share to Poste' functionality we have been discussing, by allowing a user to select
'share' from another application and share the data to Poste.

Normally, the application opens from IntroActivity, which allows user to choose to Login or Register.
This will change when a user attempts to share to Poste.

Alternate control flow:

1) User clicks 'Share' on some content and selects to share to Poste
2) Application opens NewPostActivity
3) Application checks if user is logged in.
4a) If user is logged in, proceed with post creation, returning to dashboard when done
4b) If user is not logged in, redirect to login page; when logged in, return to NewPostActivity and
proceed with post creation

New Files:
- `NewPostActivity.java`: Represents the activity to create a new post (rather than editing) -- I wanted
to work on a fresh activity rather than modifying EditPostActivity and having to touch too much of the
code in one task. Functionally very similar to EditPostActivity, but there is also a spinner (dropdown
menu) that allows a user to select the folder to create the post in. TODO: We need to determine what
happens if a User attempts to creat a Post, but has no Folders. Options: Create a default 'Home' folder,
give an error, redirect to FolderCreation, etc.
- `activity_new_post.xml`: Layout for the above activity

Changes:
- `AndroidManifest.xml`: Added intent-filter to allow sharing content to NewPostActivity
- `DashboardActivity.java`: Implemented Create Post functionality. Now, when user elects to create a post from
this page, it starts NewPostActivity, but does not finish -- this way, when post creation is finished,
user is returned here
- `LoginActivity.java`: Added checks for 'getBooleanExtra("return")'. This extra information should be passed
to `LoginActivity` when we want `LoginActivity` to simply `finish()` and return to the previous
activity when done, rather than redirecting user to the Dashboard. This is useful when a user tries to
share content to Poste, but they are not logged in -- first, it will ask them to login, but then return
them to NewPostActivity (which called Login) rather than to the Dashboard. If this extra isn't passed
to LoginActivity, they will simply be forwarded to dashboard
- `CreatePost.java`: Modified API request class to match sent data format to that expected by backend
- `Folder.java`: Implemented `toString()`, which is used to display Folder title in NewPostActivity spinner / dropdown
- `User.java`: Removed some duplicate code and unused import, and cleans up updateFoldersAndPosts
- `ValidationUtils.java`: Added some documentation

### DIFF
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -32,9 +32,6 @@ import com.example.poste.http.RetrofitClient;
 import com.example.poste.models.User;
 import com.example.poste.utils.utils;
 
-import org.json.JSONException;
-import org.json.JSONObject;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -196,7 +193,8 @@ public class DashboardActivity extends PActivity {
                         return true;
                     case R.id.menu_post:
                         // Handle post creation
-                        showCreatePostDialog();
+                        Intent intent = new Intent(DashboardActivity.this, NewPostActivity.class);
+                        startActivity(intent);
                         return true;
                     default:
                         return false;
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -9,12 +9,10 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.Window;
 import android.view.WindowManager;
-import android.widget.ArrayAdapter;
 import android.widget.Button;
 import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.PopupMenu;
-import android.widget.Spinner;
 import android.widget.Toast;
 
 import androidx.recyclerview.widget.GridLayoutManager;
@@ -27,13 +25,11 @@ import com.example.poste.callbacks.UpdateCallback;
 import com.example.poste.models.Folder;
 import com.example.poste.http.FolderRequest;
 import com.example.poste.http.MyApiService;
-import com.example.poste.http.CreatePostRequest;
 import com.example.poste.http.RetrofitClient;
 import com.example.poste.models.User;
 import com.example.poste.utils.utils;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 
 import okhttp3.ResponseBody;
@@ -50,7 +46,6 @@ public class DashboardActivity extends PActivity {
     private RecyclerView folderRecyclerView;
     private FolderAdapter folderAdapter;
     public ImageView optionsView;
-    public HashMap<String, String> folderNameToIdMap = new HashMap<>();
     private MyApiService apiService = RetrofitClient.getRetrofitInstance().create(MyApiService.class);
 
     /**
@@ -112,13 +107,6 @@ public class DashboardActivity extends PActivity {
             @Override
             public void onSuccess() {
                 userFolders = currentUser.getFolders(); // update our local copy of the folders
-                folderNameToIdMap.clear(); // clear our map of folder names to folder IDs
-                for (Folder folder : userFolders) {  // repopulate the map
-                    folderNameToIdMap.put(
-                            folder.getTitle(),
-                            folder.getId()
-                    );
-                }
                 // Fill folder view (Recycler View)
                 // Note: This is not the best way to do this, because it means every time
                 // DashboardActivity restarts, a new Adapter is created, but it works for now. The
@@ -172,7 +160,6 @@ public class DashboardActivity extends PActivity {
         // Update current user's information, which calls the updateCallback when finished
         currentUser.updateFoldersAndPosts(updateCallback);
     }
-
 
     private void showCreateItemDialog() {
         // Find the [+] button on the dashboard
@@ -286,72 +273,4 @@ public class DashboardActivity extends PActivity {
         dialog.show();
     }
 
-    // menu item select listener
-    @Override
-    public boolean onContextItemSelected(MenuItem item) {
-        Folder folder = folderAdapter.getLocalDataSetItem();
-        Intent intent = null;
-        switch (item.getItemId()) {
-            case R.id.ctx_menu_edit_folder:
-                intent = new Intent(DashboardActivity.this, EditFolderActivity.class);
-                intent.putExtra("folderId", folder.getId());
-                intent.putExtra("folderName", folder.getTitle());
-                intent.putExtra("folderShared", true);
-                startActivity(intent);
-                //finish();
-                break;
-            case R.id.ctx_menu_share_folder:
-                intent = new Intent(DashboardActivity.this, Shared_Folder.class);
-                intent.putExtra("folderId", folder.getId());
-                intent.putExtra("folderName", folder.getTitle());
-                startActivity(intent);
-                //finish();
-                break;
-            case R.id.ctx_menu_delete_folder:
-                // Delete the folder
-                Call<ResponseBody> call = apiService.deleteFolder(
-                        currentUser.getTokenHeaderString(),
-                        folder.getId()
-                );
-                call.enqueue(new Callback<ResponseBody>() {
-                    @Override
-                    public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
-                        if (response.isSuccessful()) {
-                            Toast.makeText(DashboardActivity.this, "folder deleted successful.", Toast.LENGTH_LONG).show();
-                            // Reload Dashboard
-                            Intent intent = new Intent(DashboardActivity.this, DashboardActivity.class);
-                            intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
-                            startActivity(intent);
-                            finish();
-                        } else {
-                            String error = utils.parseError(response);
-                            if (error != null) {
-                                Toast.makeText(
-                                        DashboardActivity.this,
-                                        "Folder deletion failed: " + error,
-                                        Toast.LENGTH_LONG
-                                ).show();
-                            } else {
-                                Toast.makeText(
-                                        DashboardActivity.this,
-                                        "Folder deletion failed.",
-                                        Toast.LENGTH_LONG
-                                ).show();
-                            }
-                        }
-                    }
-
-                    @Override
-                    public void onFailure(Call<ResponseBody> call, Throwable t) {
-                        Toast.makeText(
-                                DashboardActivity.this,
-                                "Folder deletion failed.",
-                                Toast.LENGTH_LONG
-                        ).show();
-                    }
-                });
-                break;
-        }
-        return true;
-    }
 }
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -91,47 +78,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+             * and that the callback was passed to it; so, what do we want to do if the update
+             * was successful?
+             */
+            @Override
+            public void onSuccess() {
+                userFolders = currentUser.getFolders(); // update our local copy of the folders
+                folderNameToIdMap.clear(); // clear our map of folder names to folder IDs
+                for (Folder folder : userFolders) {  // repopulate the map
+                    folderNameToIdMap.put(
+                            folder.getTitle(),
```
fmehdad97: Pretty sure you would want the ID as the key here, and Title/Folder as the mapped value

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -91,47 +78,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+             * This defines what will happen if the update fails; what we want the application
+             * to do in a failure case (e.g. if the API call fails).
+             * @param errorMessage The error message returned by the API call (or whatever we send
+             *                     the callback to)
+             */
+            @Override
+            public void onError(String errorMessage) {
+                Toast.makeText(
+                        DashboardActivity.this,
+                        "Unable to retrieve folders and posts, please try again.",
```
fmehdad97: Even though it is nice to get such a feedback as a user, right now the only way to correct the issue is to signout and login again (if im not wrong). Maybe we should consider implementing the refresh option we talked about. But obviously that is out of the scope of what you were working on.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -178,7 +218,11 @@ private void showCreatePostDialog() {
         Spinner spinnerNewPostFolder = dialogView.findViewById(R.id.spinnerNewPostFolder);
 
         // Setup for the dropdown menu
-        ArrayAdapter<String> adapter= new ArrayAdapter<String>(this, android.R.layout.simple_spinner_item, new ArrayList<>(folderIdNameMap.keySet()));
+        ArrayAdapter<String> adapter = new ArrayAdapter<>(
+                this,
+                android.R.layout.simple_spinner_item,
+                new ArrayList<>(folderNameToIdMap.keySet())
```
fmehdad97: I was unable to get a dropdown when testing this. Could you please check again that the dropdown works when creating a post? And without a folder selected, cannot create a post either.

fmehdad97: image
This is what I get in log when I try to get list of folders

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -195,30 +239,54 @@ public void onClick(DialogInterface dialog, int which) {
...
+                if (itemName == "" || itemLink == null || itemLink == "") {
+                    Toast.makeText(
+                            DashboardActivity.this,
+                            "Cannot create post, please enter name and link",
+                            Toast.LENGTH_LONG
+                    ).show();
                     return;
                 }
 
                 if (!itemLink.matches("^((http|https):\\/\\/)(www.)?[a-zA-Z0-9@:%._\\\\+~#?&/=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%._\\\\+~#?&/=]*)$")) {
```
fmehdad97: (Out of scope) Might be really helpful to move this to utilility functions.

fmehdad97: Although, if I'm not wrong this only accepts URLs that only begin with www. So some links might incorrectly deemed invalid. For example, when sharing from different apps some apps do it weird, such as facebook might do a m.facebook.com/..... or certain ones can do app.xyz.com/....

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -195,30 +239,54 @@ public void onClick(DialogInterface dialog, int which) {
...
 
                 call.enqueue(new Callback<ResponseBody>() {
                     @Override
                     public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
-                        if(response.isSuccessful()) {
-                                    // create a local post matching the one created on the API
-                                    currentUser.updateFoldersAndPosts(DashboardActivity.this);
-                                    Toast.makeText(DashboardActivity.this, "Post creation successful.", Toast.LENGTH_LONG).show();
+                        if (response.isSuccessful()) {
+                            // Display success message, then reload Dashboard
```
fmehdad97: Given the fact that dashboard activity updates folder and post anyways, I think this is great for simplicity

app/src/main/java/com/example/poste/activities/NewPostActivity.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_new_post);
+        if (User.getUser().getToken() == null || User.getUser().getToken().isEmpty()) {
+            // user needs to login
+            Log.d("NewPostActivity", "User needs to login");
+            Intent loginIntent = new Intent(this, LoginActivity.class);
+            loginIntent.putExtra("return", true);
```
fmehdad97, line 68: So once user logs in, the app finishes that activity, and since this activity never finished, it bring it back here? 
Am I correct? 

Only concern I have is that, the login (or maybe dashboard) would update the user data such as folder and posts. And in this control flow I don't really see the update happening, which might be why I am unable to view the dropdown folders I mentioned in a previous comment

app/src/main/java/com/example/poste/http/MyApiService.java
```diff
@@ -46,10 +46,10 @@ Call<ResponseBody> createFolder(
             @Body FolderRequest folderRequest
     );
 
-    @GET("folders/folders/deleteFolder/{id}/")
+    @DELETE("folders/{id}/")
```
fmehdad97: Thank you for fixing this method and making it more REST-ful!

app/src/main/java/com/example/poste/http/PostRequest.java
```diff
@@ -4,9 +4,9 @@ public class PostRequest {
     private String title;
     private String description;
     private String url;
-    private int folder;
+    private String folder;
 
-    public PostRequest(String title, String description, String url, int folder) {
+    public PostRequest(String title, String description, String url, String folder) {
```
fmehdad97: You mean String folderID?

fmehdad97 (changes requested):
Most of the comments on review are non-blocking.

The only major change requested is the fact that in my run of the application, I was unable to get the folders drop down to work

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: I only moved (after renaming) the mapping; can you elaborate? 

This populates the map with title as key and id as values, and later populates the folder names by key and identifies them by their id. Maybe there's something I'm missing?

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: Could hit back, or a drag-down, yeah. Good input; might be valuable to create a User Story out of that though so it doesn't get lost in this PR

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: Sure -- we might want to just remove URL validator altogether. If a User wants to enter their ex-girlfriend's phone number instead of a URL, I feel that's on them and we shouldn't really stop them.

(Actually, this creates an interesting domain for improvements -- I'm pretty sure that opening a link with a phone number may just try to call someone; could we make this a one-stop-shop for saving *everything*?)

app/src/main/java/com/example/poste/activities/NewPostActivity.java
mccodav1: Once user logs back in, LoginActivity (having received this 'extra' of return) knows to return to NewPostActivity (instead of Dashboard) to finish allowing the user to create a post, correct.

You're correct about data updating -- I'll add data updating to LoginActivity in this case (since they won't hit dashboard first)

The dropdown folders aren't fully implemented yet. I considered that part of EditPostActivity... although now I guess this is a whole new activity (this branch evolved over time) so I should probably create it here. Then, it can be shared in EditPostActivity, if we so desire.

app/src/main/java/com/example/poste/http/PostRequest.java
mccodav1: I only changed the type here so we weren't casting to and from strings and ints; this should match what the backend expects to receive though, which appears to be `folder_id`. Just had to scope out what I was covering in one PR

mccodav1: I have refactored the name to `folder_id` as this is what the serializer on the backend expects, I believe.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
fmehdad97: The title may or may not be unique. So if 2 folders have the same title, the second would overwrite the first and the app would only show one folder.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
fmehdad97: For now, removing the validator altogether seems acceptable to me. 
but opening up the domain to anything and everything might be _slightly_ out of scope given that the layouts and everything else explicitly ask the user to put down a URL.

mccodav1:
> Most of the comments on review are non-blocking.
> 
> The only major change requested is the fact that in my run of the application, I was unable to get the folders drop down to work

Fixed.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: I hadn't implemented that yet, but it is implemented now.

app/src/main/java/com/example/poste/activities/NewPostActivity.java
mccodav1: I added data updating in the appropriate location in NewPostActivity to ensure data is retrieved before trying to do anything fancy

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: Oh, you right. Good looking out. It was originally created as a Hashmap (don't recall who created it) and I just changed the name) but was only considering it as a list. Makes sense. ID to Title makes perfect sense there, then.

mccodav1: Actually, I just realized that I abstracted out the usage of the map altogether -- that's why I didn't notice earlier. Good eye though! I will be removing this completely though as it is no longer used.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: No longer used since we moved to NewPostActivity altogether

dzemlin (approved):
Looks good. Please merge as soon as you can

mccodav1:
> Looks good. Please merge as soon as you can

I need @fmehdad97 to approve the changes he requested; it blocks me from merging.

fmehdad97 (approved):

fmehdad97:
Tested and found to be working. Great work!

PR #19, opened 2023-11-06
added checkstyle to the project and a gradle task to run it.

atreichl:


atreichl:
This is the checkstyle i used: https://github.com/noveogroup/android-check/tree/master

mccodav1:
I like it. There's a lot of items that are marked as ignored or have lowered severity; do we not want to check for those items?

Is it possible to include this into CICD?

mccodav1:
> This is the checkstyle i used: https://github.com/noveogroup/android-check/tree/master

This is cool; is there a way to utilize the whole plugin, rather than just the checkstyle config?

atreichl:
I made the check style harder there was an easy and hard checkstyle.  I had to edit the checkstyle.xml for it to work, so I don't know if we can use the plugin as a whole, but I am going to added the spotbugs from it as well. I will look into if we can add it to the CICD checks.

mccodav1 (approved):
Looks good. I think we should find a way to integrate this in CICD to ensure it gets run; however, I'm mindful that including this as-is now would likely fail our builds for a while until we can sort them all out... I wonder if its best to implement this as a non-failing CICD job?

PR #20, opened 2023-11-07
US 128 MVP activity flow unification and Bug documentation

dzemlin:
Bugs added to backlog (and some are in sprint 4 now).

1 critical bug needs to be fixed before  MVP is released to Nadia ( https://tree.taiga.io/project/dzemlin-poste/us/202?milestone=368325 )
Edit Post does not successfully update posts. We just get a failed toast message. I suspect the API call is not formed correctly.

As for the flow of the menus. only major issue is the page stacking bug. This is already User story in sprint 4 which I have begun working on.


app/src/main/java/com/example/poste/activities/EditFolderActivity_v2.java
```diff
@@ -103,5 +102,6 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
     private void sendToDashboard() {
         Intent newIntent = new Intent(EditFolderActivity_v2.this, DashboardActivity.class);
         startActivity(newIntent);
+        finish();
```
mccodav1, line 105: We may be able to just call finish() without creating and starting a new intent; this would return the user to where they were before. While this is likely the dashboard anyway, I think it could be beneficial to prevent stacking up of more activity frames

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -57,9 +60,6 @@ protected void onCreate(Bundle savedInstanceState) {
         TextView folderName = findViewById(R.id.folderNameText);
         ImageButton newBut = findViewById(R.id.newPost);
         ImageButton settingsBut = findViewById(R.id.folderSettings);
-        Button openBut = findViewById(R.id.buttonOpen);
```
mccodav1, line 60: Thanks -- I think clicking / long clicking makes more sense

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,47 +72,10 @@ protected void onCreate(Bundle savedInstanceState) {
 
         settingsBut.setOnClickListener(view -> {
             Intent intent = new Intent(FolderViewActivity.this, EditFolderActivity_v2.class);
+            intent.putExtra("folderId", PosteApplication.getSelectedFolder().getId());
```
mccodav1, line 75: If we are setting PosteApplication.selectedFolder, could we just pull from that field in EditFolder activity rather than sending / retrieving extras on the intent?

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -72,47 +72,10 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                    }
-
-                    @Override
-                    public void onError(String errorMessage) {
-                        Toast.makeText(FolderViewActivity.this, errorMessage, Toast.LENGTH_LONG).show();
-                    }
-                };
-                User.getUser().deletePostFromServer(PosteApplication.getSelectedPost(), postDeletionCallback);
-            }
+            finish();
```
mccodav1, line 78: It might be best not to call finish() here so when the user is done editing, they are returned to this folder -- thoughts?

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -135,15 +93,62 @@ public void onError(String errorMessage) {
...
+                            startActivity(intent);
+                        }
+
+                        @Override
+                        public void onError(String errorMessage) {
+                            Toast.makeText(FolderViewActivity.this, errorMessage, Toast.LENGTH_LONG).show();
+                        }
+                    };
+                    User.getUser().deletePostFromServer(PosteApplication.getSelectedPost(), postDeletionCallback);
+                    finish();
```
mccodav1, line 147: I think this finish() and recreate() should be removed; we don't want them to run before the callbacks or we will have unexpected behavior

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -135,15 +93,62 @@ public void onError(String errorMessage) {
...
+                break;
+            case R.id.ctx_menu_delete_post:
+                PosteApplication.setSelectedPost(post);
+                if (PosteApplication.getSelectedPost() != null) {
+                    PostDeletionCallback postDeletionCallback = new PostDeletionCallback() {
+                        @Override
+                        public void onSuccess() {
+                            Toast.makeText(FolderViewActivity.this, R.string.post_deleted, Toast.LENGTH_LONG).show();
+                            Intent intent = new Intent(FolderViewActivity.this, FolderViewActivity.class);
+                            finish();
```
mccodav1, line 137: instead of creating a new intent, calling finish() , and then starting it again, a recreate() serves a similar purpose (but refreshes faster, it appears, at least locally)

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -135,15 +93,62 @@ public void onError(String errorMessage) {
...
+                            startActivity(intent);
+                        }
+
+                        @Override
+                        public void onError(String errorMessage) {
+                            Toast.makeText(FolderViewActivity.this, errorMessage, Toast.LENGTH_LONG).show();
```
mccodav1, line 143: if removing finish() and recreate() from below the callbacks, a recreate() here would be appropriate

app/src/main/java/com/example/poste/activities/OptionsActivity.java
```diff
@@ -75,7 +63,5 @@ public void onComplete(Task<Void> task) {
                 startActivity(new Intent(OptionsActivity.this, LoginActivity.class));
```
mccodav1, line 63: it would be beneficial to create a quick method that actually logs the user out instead of just moving them to login page; otherwise, simply pressing 'back' would be sufficient to steal the user session.

A User.logout() method would simply set all user fields to null (the token is particularly important)

mccodav1 (approved):
Conditional approval, with the caveat that the items marked should either be changed before merging or new user stories created with specific details and perhaps included in the *current* sprint, as they affect functionality I think is important to our demo.

app/src/main/java/com/example/poste/activities/OptionsActivity.java
fmehdad97: I agree with @mccodav1 
Additionally, might be slightly overkill, if an API call was made to the server, the server could invalidate the token it had generated for the user, making the application more secure as lingering tokens are a big security issue in most web applications.

mccodav1: 100%, and not overkill

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Yes. agreed. I saw 2 different way of doing this in our code.
I have made a consistency US ( https://tree.taiga.io/project/dzemlin-poste/us/217?no-milestone=1 )

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: Good catch I can address that in my ( https://tree.taiga.io/project/dzemlin-poste/us/185?milestone=368325 )

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: ok I was wondering if Recreate() would be useful to us here. 
fits nicely into https://tree.taiga.io/project/dzemlin-poste/us/185?milestone=368325

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
dzemlin: understood

app/src/main/java/com/example/poste/activities/OptionsActivity.java
dzemlin: Created ne US for this:
https://tree.taiga.io/project/dzemlin-poste/us/221?no-milestone=1

PR #22, opened 2023-11-10
Us189 Change Password 

fmehdad97:
To test:
Login -> Click gear icon on dashboard -> Click Account Settings -> enter old password, new password and confirm new password. 

Backend server URL implemention on branch: 
https://github.com/Appyo-Poste/PosteBackend/tree/us189_changePassword_fmehdad


app/src/main/java/com/example/poste/utils/ValidationUtils.java
```diff
@@ -55,11 +55,17 @@ public static boolean validateEmail(Context context, String email) {
...
+                Toast.makeText(context, "Passwords do not match!", Toast.LENGTH_SHORT).show();
+                return false;
+            } else {
+                return true;
+            }
+        } else{
             return false;
         }
-        return validatePassword(context, password);
+
```
mccodav1, line 68: Can you elaborate on the change here?

app/src/main/java/com/example/poste/http/EditAccountRequest.java
```diff
@@ -0,0 +1,11 @@
+package com.example.poste.http;
+
+public class EditAccountRequest {
+    private String oldPassword;
+    private String newPassword;
+
+    public EditAccountRequest(String oldPassword, String newPassword) {
```
mccodav1, line 7: I'm wondering if we should maybe rename this to be more precise, or alternatively allow passing a key-value listing such that we can use this same endpoint for editing different account fields in the future?

mccodav1 (approved):

PR #23, opened 2023-11-10
Updated dashboard UI as best I can.

atreichl:
Could not get the folder behind the add button also with the add button adding more then just folders does not make much sense at the moment anyway.


fmehdad97 (approved):
Tested and looks pretty good.

We do however need to standardize the size of the top bar and the logo for uniformity across activities. Something to keep in mind for future sprints though.

PR #24, opened 2023-11-10
Us192t233

mccodav1:
This updates NewPostActivity to match Prototype. Might be useful as a reference for others wanting to match the overlapping Poste icon, colors, button schemes, etc. Built off of @atreichl 's UI improvement branch


fmehdad97 (approved):
Tested and working as expected.

Only one comment: It would be very nice if it was possible to give any kind of visual cue to the user that select Folder is a dropdown.

mccodav1:
> Tested and working as expected.
> 
> Only one comment: It would be very nice if it was possible to give any kind of visual cue to the user that select Folder is a dropdown.

Low priority at the moment I'm afraid lol. The easiest way would be for the first option in the spinner to simply be --Click to select a folder-- or something.

PR #26, opened 2023-11-11
Fix control flow to exit from EditPostActivity

fmehdad97:


app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -113,6 +112,7 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
                             currentPost.setDescription(description);
                             currentPost.setUrl(url);
                             Toast.makeText(EditPostActivity.this,"Edit post successful!", Toast.LENGTH_SHORT).show();
+                            finish();
```
mccodav1, line 115: Would suggest placing this after the if/else for response.isSuccessful, that way it returns on either case; onFailure as well

mccodav1, line 115: Would suggest placing this after the if/else for response.isSuccessful, that way it returns on either case; onFailure as well

mccodav1 (changes requested):
Positive change, but I think we could relocate the finish() so it occurs in any case. Also, instead of commenting those lines out, maybe we should remove them completely?

app/src/main/java/com/example/poste/activities/EditPostActivity.java
fmehdad97: So my thought for using the finish within the successful block only was that if the edit failed for some reason and the user would like to try again. If we finish even in case of a failure, the changes that the user made will be lost. 
The control flow in my head was that if the edit failed, and the user would like to exit, they could click cancel changes and that would finish the activity.

As for the commented lines, I can go ahead and remove them to reduce confusion.

mccodav1: To avoid merge conflict with the big US185 branch, is it acceptable to you to include these changes in the US185 PR and close this one? You  make a good point on not calling finish() on failure cases and I'll implement that there.

fmehdad97: I see you have included it already. So I will go ahead and close out this PR.

PR #27, opened 2023-11-11
Match editPost to match the prototype layout as much as possible.

fmehdad97:
Based off of the layout designed by @mccodav1 for create new post.


mccodav1 (approved):
Looks great -- only suggestion I have would be for us to come to an agreement on how the title / icon look (size, placement) -- ideally these two pages (NewPost and EditPost) should match as much as possible

fmehdad97:
I'll make a task to improve upon your suggestion.

PR #28, opened 2023-11-11
US185 - Control Flow Revisions

mccodav1:
Built from @dzemlin 's US185 branch and includes some revisions to squash the listed bugs and (hopefully) finalize a functional, (mostly) bug-free control flow throughout the application.

Demo video, since this is a big PR: https://youtu.be/HQK00DcYZXQ

I also walked through these changes with Fayez so there was some visibility and pair-programming. Thanks for the hand, @fmehdad97 !

- Attempts to clean up finish() calling: We want to make sure finish() is called when we don't want a user to return to a given activity, and that we only start a new intent and send a user there when we aren't able to just return a user back to their previous screen. Priority should be returning the user to where they were except in specific cases (for instance, when a user wants to create a post, we have to go somewhere. When they're done, they should likely just be returned back.)
- Cleans up Dashboard code
- Adds a drag-down-to-refresh on Dashboard
- Refactors some Activity / Layout names for consistency
- Various refactors / cleanups / method extraction


dzemlin (approved):
Approved!

PR #29, opened 2023-11-11
UI changes for Demo

fmehdad97:
The following activities were changes to match the prototype as closely as possible
1. Edit Post
2. Intro Activity
3. Login
4. Register (still missing the hyperlink on Register that takes a user to login screen "Already have an account? Login"
5. Account Options (partially updated. Missing gray background editText)
6. Share Folder
7. Edit Folder

N.B. a few did not have a prototype design, so they were improvised to match the theme as best as possible.


dzemlin (approved):
Looks Great!
All menus look like they belong to the same app now!

PR #30, opened 2023-11-13
Optimize imports and remove duplicate code

fmehdad97:
Small changes

- Removed duplicate CreatePost request
- Optimized imports; remove unused imports and group imports together by package
- Small UI change to make folder view top bar look identical to dashboard
- editPost no longer required FolderID to find post. Only Post ID needed. Removed unused functions too.


app/src/main/java/com/example/poste/PosteApplication.java
```diff
@@ -49,7 +42,7 @@ public static com.example.poste.api.poste.models.User getCurrentUser() {
      *
      * @param currentUser The User object representing the currently logged-in user
      */
-    public static void setCurrentUser(com.example.poste.api.poste.models.User currentUser) {
+    public static void setCurrentUser(User currentUser) {
```
mccodav1: This can (and should) go altogether. Since we are using Singleton, we want to avoid storing a user anywhere as much as possible (in case user logs out or whatnot) and instead just use User.getUser()

mccodav1 (commented):
Looks good, but I think we could take it a step further and remove PosteApplication altogether. After reviewing, it seems the only methods still in use involve actions User could perform (getUser(), setting active folder / post, etc) -- can we remove this here?

mccodav1 (approved):

PR #31, opened 2023-11-14
Add additional optimization suggestions

mccodav1:
Added additional refactoring / removal of unused assets

I replaced PosteApplication with User methods, and ripped out a bunch of unused stuff. Have tested the application working but please double check things work as intended.


fmehdad97 (approved):
Tested and found to be working as expected.

Thank you for the trim down of codebase

PR #33, opened 2023-11-18
Us 194 front end can send tags

dzemlin:


dzemlin:
*Bug fixed*
Ready to be reviewed

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -71,6 +72,16 @@ protected void onCreate(Bundle savedInstanceState) {
             postTitle.setText(currentPost.getTitle());
             postDescription.setText(currentPost.getDescription());
             postLink.setText(currentPost.getUrl());
+            if (!currentPost.getTags().isEmpty()) {
+                String tagsNewText = "";
+                for (String s : currentPost.getTags()) {
+                    if (tagsNewText.length() > 0) {
+                        tagsNewText = tagsNewText + ",";
```
fmehdad97: From a UI POV, have a space after a comma is much preferred

app/src/main/java/com/example/poste/models/User.java
```diff
@@ -209,13 +209,29 @@ public void onResponse(@NonNull Call<ResponseBody> call, @NonNull Response<Respo
                                 String postDescription = post.getString("description");
                                 String postUrl = post.getString("url");
                                 String postId = post.getString("id");
+                                String tagList = "";
+                                if (post.has("tags")) {
+                                    JSONArray postTags = post.getJSONArray("tags");
+                                    for (int p = 0; p < postTags.length(); p++) {
+                                        tagList = tagList + postTags.getString(p);
+                                        if (p < postTags.length() - 1) {
+                                            tagList = tagList + ",";
```
fmehdad97: I don't think it is a good idea to conver the JSON Array into a String, and then pass that string into the setTags method to convert it back into a list.
You cna try the following
https://stackoverflow.com/questions/3395729/convert-json-array-to-normal-java-list

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -27,6 +31,14 @@ public void setUrl(String url) {
         this.url = url;
     }
 
+    public void setTags(String tags) {
+        if (tags.length() < 1) {
+            return;
+        }
+        ArrayList<String> newTagList = new ArrayList<String>(Arrays.asList(tags.split(",")));
```
fmehdad97: See the comment on User class.
It can be as simple as setTags(List<String> tags)

fmehdad97 (changes requested):
https://www.loom.com/share/f3586a5add59460495f4c1abe830d17f?sid=00ddf11f-30b1-40b2-9746-82aa68d988a6

EditPost functionality not working for me. While the tags saved when creating a post does show, if I add a new tag on edit, it does not seem to be working. Tried on entities that had one or more tags and also ones that did not have any tags

app/src/main/java/com/example/poste/activities/EditPostActivity.java
```diff
@@ -71,6 +72,16 @@ protected void onCreate(Bundle savedInstanceState) {
             postTitle.setText(currentPost.getTitle());
             postDescription.setText(currentPost.getDescription());
             postLink.setText(currentPost.getUrl());
+            if (!currentPost.getTags().isEmpty()) {
```
mccodav1: I think this can actually be dropped; if we're just iterating over the tags, an empty tag list will just end right away

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -47,6 +64,7 @@ private Post(Builder builder) {
         this.description = builder.description;
         this.url = builder.url;
         this.id = builder.id;
+        this.tags = builder.parseTags();
```
mccodav1: This can cause a NullPointerException in other locations; make sure we are checking for that when retrieving tags. (Really, the same can happen for the other fields, except they're required on creation so that can't really happen. With tags, they CAN be empty, in which case they won't be set... and then retrieving / iterating over tags will throw that exception)

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -79,6 +97,7 @@ public String getUrl() {
      */
     public String getId() { return id; }
 
+    public ArrayList<String> getTags() {return tags; }
```
mccodav1: Considering the NPE risk mentioned above, I'd suggest we place some logic here instead of just returning tags. I would suggest a return type of List (for polymorphism) and then we can create a new empty list (ArrayList works) and add all tags to it; if tags is null, its returned an empty list. That would help avoid the NPE risk.

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -104,6 +123,11 @@ public static class Builder {
          */
         private String id;
 
+        /**
+         * The list of tags as one comma separated string.
+         */
+        private String tags;
```
mccodav1: Might be worth making a US (or working on the backend side-by-side in this US) to change over from a comma-separated string to an actual list, which would help alleviate some odd logic we have to use here in the front end

app/src/main/java/com/example/poste/activities/EditPostActivity.java
dzemlin: Change to use ", "

app/src/main/java/com/example/poste/activities/EditPostActivity.java
dzemlin: That's some AM-hours coding side affects there.
Removed.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Fixed.
 now you will always get get/set and array lists. the list can be empty, but wont be null

dzemlin:
> https://www.loom.com/share/f3586a5add59460495f4c1abe830d17f?sid=00ddf11f-30b1-40b2-9746-82aa68d988a6
> 
> EditPost functionality not working for me. While the tags saved when creating a post does show, if I add a new tag on edit, it does not seem to be working. Tried on entities that had one or more tags and also ones that did not have any tags

Fixed all front end issues that were found
I will start working on the back end implementation of Edit Post, so that is understands and adds tags. 
For now though, this branch I think is good to pull, so that others can start working on tag related features (like searching Via tags)

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Good point. 
Now it creates a copy of the list and returns it.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Agreed
I think that is a point worth discussing next time we meet.

app/src/main/java/com/example/poste/models/Post.java
dzemlin: Done.

app/src/main/java/com/example/poste/models/User.java
dzemlin: Changed. We now just go strait from json array to arraylist which is given to the builder.

fmehdad97 (approved):
Approving conditionally as the editing tags non-working status has been diagnosed to be a backend issue.

PR #34, opened 2023-11-20
Task 256 - App using https for comuncation

atreichl:
You need to be on the back end branch us191t260 to test. This should not be merged till that branch is ready to be merged into the back end. us191t260 has info about generating certs and setting them on the back end so follow those first then follow instruction in the readme of this branch to set the cert on the app side. This is a video showing setting it up and working on my end https://youtu.be/6rvUS92sPUk. The long failed due to me mistyping the password.

### DIFF
--- a/app/src/main/java/com/example/poste/http/RetrofitClient.java
+++ b/app/src/main/java/com/example/poste/http/RetrofitClient.java
@@ -2,6 +2,7 @@ package com.example.poste.http;
 
 import com.example.poste.R;
 import com.example.poste.models.PosteApp;
+import com.example.poste.utils.NullHostNameVerifier;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -14,6 +15,7 @@ import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/example/poste/http/RetrofitClient.java
```diff
@@ -1,27 +1,131 @@
...
+
 import okhttp3.OkHttpClient;
 import okhttp3.logging.HttpLoggingInterceptor;
 import retrofit2.Retrofit;
 import retrofit2.converter.gson.GsonConverterFactory;
 
 public class RetrofitClient {
 
-    private static final String BASE_URL = "http://10.0.2.2:8000/api/";
+    private static final String BASE_URL = "https://10.0.2.2:443/api/";
```
mccodav1: The :443 can be removed; https automatically uses port 443

mccodav1 (commented):
Appears to be functional, great job.

One concern that I have is that this appears to nullify hostname checking, which would make the application susceptible to man-in-the-middle attacks. Can you make a US to investigate / solve this down the road? It is something we will definitely not want to ship in the final product, as it would be considered a major security vulnerability, but the backend has similar security improvements that will be necessary (such as setting Django ALLOWED_HOSTS and CSRF_TRUSTED_ORIGINS, among other things)

I think this suffices for the given task, however.

I'll test this against my amazon server and verify it continues to work

mccodav1 (approved):

mccodav1:
One other thing -- can you add '*.crt' and '*.key' to .gitignore? You can also remove 'poste.txt' and replace it with a blank file named simply '.gitkeep' (mind the period) which will be invisible but tells git to track the directory

mccodav1 (changes requested):
One small thing -- can you add '.crt' and '.key' to .gitignore? You can also remove 'poste.txt' and replace it with a blank file named simply '.gitkeep' (mind the period) which will be invisible but tells git to track the directory

app/src/main/java/com/example/poste/http/RetrofitClient.java
atreichl: it has been removed.

atreichl:
It need the placeholder file with the same name otherwise the build will fail.

atreichl:
I will look into host name verification, it unfortunately does not seem to be a easy fix or simply fix.

mccodav1:
> I will look into host name verification, it unfortunately does not seem to be a easy fix or simply fix.

Understood. It doesn't have to happen now, but it does have to happen. MITM attacks are quite common, and without verifying the hostname, any cert can be used in place of our own. I think it would be more prudent to release the product with HTTP only (despite its shortcomings) rather than obfuscate a security risk of this size.

That being said, for this stage of development, I think this is acceptable, but only if we have a user story to come back to this

atreichl:
I have created a user story for it all ready and have started looking into it, I should have it sorted out tomorrow morning.

mccodav1 (commented):
For context, the backend MR should be reviewed + merged before this is: https://github.com/Appyo-Poste/PosteBackend/pull/28

Otherwise, this MR appears good to go.

mccodav1 (approved):

PR #36, opened 2023-11-23
Abstract API endpoint to build variable

mccodav1:
This commit abstracts the API endpoint to a build variable, which allows us to dynamically switch between using a local server or the development server.

The development server is located at: https://postebackend.duckdns.org

This adds a new build config; there are now four:
- `local`: The API endpoint will be set to https://10.0.2.2/api/ (uses a local backend)
- `development`: The API endpoint will be set to  https://postebackend.duckdns.org/api/ (Amazon EC2 development server)
- release: Previously existing. Did not change.
- debug: Previously existing. Did not change

To use a different build variant, do the following:

1) Open Android studio
2) In Menu bar, click Build > Select Build Variant
3) In the Build Variant tool window, change between `development` or `local` as desired.

When changing Build Variant, you may need to rebuild the project manually:
1) In the Menu bar, click Build > Clean Project
2) In the Menu bar, click Build > Rebuild Project
3) Run application

Occasionally, after changing build variant, running the application may throw an error:
`Couldn't delete Poste\app\build\intermediates\compile_and_runtime_not_namespaced_r_class_jar\development\R.jar`

Clicking Run again should remove the error as it rebuilds; this may also be resolved by changing the Run configuration.

NOTE: `Poste.crt` must be present in `res/raw` since this now uses HTTPS; see the Backend repo for crt generation instructions.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: set up JDK 11
+    - name: set up JDK 17
       uses: actions/setup-java@v3
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'temurin'
         cache: gradle
 
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -16,11 +16,11 @@ jobs:
     - name: set up JDK 11
       uses: actions/setup-java@v3
       with:
-        java-version: '17'
+        java-version: '11'
         distribution: 'temurin'
         cache: gradle
 
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Build with Gradle
-      run: ./gradlew build
+      run: ./gradlew assembleDevelopment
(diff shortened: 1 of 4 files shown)


atreichl (approved):
The android code looks good to me I was able to connect to both my local and the AWS back ends. The certificate that you generated will not function once PR38 is merged in since the subject alt names of the certificate is empty.

To generate a certificate that would work for both local and the server once PR38 is merged in run:
openssl req -x509 -newkey rsa:4096 -keyout poste.key -out poste.crt -days 365 -nodes -addext "subjectAltName = IP:10.0.2.2,DNS:postebackend.duckdns.org"

mccodav1:
> The android code looks good to me I was able to connect to both my local and the AWS back ends. The certificate that you generated will not function once PR38 is merged in since the subject alt names of the certificate is empty.
> 
> To generate a certificate that would work for both local and the server once PR38 is merged in run: openssl req -x509 -newkey rsa:4096 -keyout poste.key -out poste.crt -days 365 -nodes -addext "subjectAltName = IP:10.0.2.2,DNS:postebackend.duckdns.org"

I have a backend MR up that has a san.cnf file to auto generate appropriate certs, too

PR #38, opened 2023-11-23
got host name verifaction working.

atreichl:
needs subject alt names in the certificate.


app/src/main/java/com/example/poste/http/RetrofitClient.java
```diff
@@ -39,8 +38,7 @@ public static Retrofit getRetrofitInstance() {
             logging.setLevel(HttpLoggingInterceptor.Level.BODY);
             client = new OkHttpClient.Builder()
                     .readTimeout(5, TimeUnit.SECONDS)
-                    .addInterceptor(logging)
-                    .hostnameVerifier(new NullHostNameVerifier());
+                    .addInterceptor(logging);
```
mccodav1, line 41: I'm not familiar, but it looks like this doesn't use a host name verifier anymore?

atreichl: okhttp has a default on it uses, which does what we need it to do. If you run it with a certificate that does not have any subject alt names it will not connect, once you add the host to subject alt names it will connect.

mccodav1 (approved):

PR #39, opened 2023-11-23
Remove nullhostverifier

mccodav1:
Having fixed the cert issue, I have removed the null host verifier.

I have tested this against our development server using branch `us270t300`. I have verified the certs are all matching. Go to https://drive.google.com/drive/u/1/folders/1j98_tcubzJN6FDu242OCpX6qWqD4A0L1 and follow the READMEs for the certs and more instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,6 @@ local.properties
 apikey.properties
 apikey.properties
 .idea/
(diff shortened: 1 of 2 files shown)


atreichl:
I can also connect to localhost and and the development server.

PR #41, opened 2023-11-25
Us 253 fix back from register

dzemlin:
Simple fix of the register activity.
Now the intro remains open so that when the register activity backs out, it will return to the intro activity.

### DIFF
--- a/app/src/main/java/com/example/poste/activities/RegisterActivity.java
+++ b/app/src/main/java/com/example/poste/activities/RegisterActivity.java
@@ -95,7 +95,7 @@ public class RegisterActivity extends AppCompatActivity {
         });
         backButton.setOnClickListener(v -> {
             // Simulate a back press; go back to LandingActivity
-            finish();
+            onBackPressed();
         });
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/example/poste/activities/RegisterActivity.java
```diff
@@ -95,8 +95,6 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
         });
         backButton.setOnClickListener(v -> {
```
mccodav1, line 96: This method can simply call `onBackPressed()`, which is the behavior if a user presses the phone's back button. It automatically calls finish(), and, since we might later modify that behavior, would be most consistent to ensure we call it here

mccodav1 (changes requested):

app/src/main/java/com/example/poste/activities/RegisterActivity.java
dzemlin: Done.

mccodav1 (approved):

PR #43, opened 2023-11-26
Enable sharing / unsharing / remove 'No access' from share

mccodav1:
This modifies the front end to parse the shared users from the backend (implemented here: https://github.com/Appyo-Poste/PosteBackend/pull/33 )

With these two branches, we now receive users shared with and can unshare from the "View Shares" context option from the Dashboard.


atreichl (approved):
Looks good to me, was able to share and then unshare a folder with a diffrent user.

PR #44, opened 2023-11-26
Remember me

mccodav1:


atreichl (changes requested):
The remember me functionality works. I will approve once you add back the placeholder poste.txt file so the GitHub build does not fail.

mccodav1:
> The remember me functionality works. I will approve once you add back the placeholder poste.txt file so the GitHub build does not fail.

We need to remove poste.txt from tracking somehow, this is going to be a pain through semester two. Just noticed this, pushing it up now.

atreichl (approved):
looks good to go now. could we not added poste.txt to .gitignore so that local changes to it will be ignored?

fmehdad97:
image

So i came across this while testing other things. 
If you clicked remember me, and then deleted the user, or the token, or the database (in my case I deleted the database), the user no longer exists but the application keeps you logged into the ghost account and shown above is the log that it returns. I think it might be beneficial to add that check in, especially as we start thinking about the multiplatform aspect of the application

PR #45, opened 2023-11-28
abstracted strings to string.xml

atreichl:
I did not abstract log strings, also android studio shows the string referenced over the reference at times. I have checked all the screens and as many of the popups as I could they are all correct for me.


app/src/main/java/com/example/poste/activities/AccountActivity.java
```diff
@@ -112,14 +112,14 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
...
-                            Toast.makeText(AccountActivity.this, "Password change failed, unknown error", Toast.LENGTH_SHORT).show();
+                            Toast.makeText(AccountActivity.this, getString(R.string.password_change_failure), Toast.LENGTH_SHORT).show();
                         }
                     }
                 }
 
                 @Override
                 public void onFailure(Call<ResponseBody> call, Throwable t) {
-                    Toast.makeText(AccountActivity.this, "Edit failed, unknown error", Toast.LENGTH_SHORT).show();
+                    Toast.makeText(AccountActivity.this, getString(R.string.edit_failure), Toast.LENGTH_SHORT).show();
```
mccodav1, line 121: Should this not match the messages from above? The ones above mention passwords, but this one just says "Edit failed"

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -265,10 +265,10 @@ private void showCreateFolderDialog() {
         // Create the AlertDialog
         AlertDialog.Builder builder = new AlertDialog.Builder(this);
         builder.setView(dialogView);
-        builder.setTitle("Create Folder");
+        builder.setTitle(getString(R.string.new_folder));
```
mccodav1: suggest `create_folder` for consistency

app/src/main/java/com/example/poste/activities/NewPostActivity.java
```diff
@@ -134,7 +134,9 @@ private void handleNewPost() {
             String folder_id = ((Folder) chooseFolderSpinner.getSelectedItem()).getId();
             if (folder_id.equals("0")) {
                 Log.d("NewPostActivity", "Default folder (0) selected");
-                Toast.makeText(this, "Select a folder to save post!", Toast.LENGTH_SHORT).show();
+                Toast.makeText(this,
+                        getString(R.string.pick_a_folder),
```
mccodav1: This could be named a bit more descriptively, so reading through strings indicates its place of usage

app/src/main/java/com/example/poste/utils/ValidationUtils.java
```diff
@@ -17,16 +19,24 @@ public class ValidationUtils {
...
         } else if (lastName.isEmpty()) {
-            Toast.makeText(context, "Please enter your last name!", Toast.LENGTH_SHORT).show();
+            Toast.makeText(context,
+                    R.string.enter_last_name,
+                    Toast.LENGTH_SHORT).show();
             return false;
         } else if (firstName.length() > 32) {
-            Toast.makeText(context, "First name must be at most 32 characters!", Toast.LENGTH_SHORT).show();
+            Toast.makeText(context,
+                    R.string.first_name_length,
```
mccodav1: Suggest name like "error_first_name_length" so it is clear this is an error message of some type (as opposed to just a length)

app/src/main/java/com/example/poste/utils/ValidationUtils.java
```diff
@@ -17,16 +19,24 @@ public class ValidationUtils {
...
+                    Toast.LENGTH_SHORT).show();
             return true;
         } else if (lastName.length() > 32) {
-            Toast.makeText(context, "Last name must be at most 32 characters!", Toast.LENGTH_SHORT).show();
+            Toast.makeText(context,
+                    R.string.last_name_length,
```
mccodav1: Suggest name like "error_last_name_length" so it is clear this is an error message of some type (as opposed to just a length)

app/src/main/java/com/example/poste/utils/ValidationUtils.java
```diff
@@ -76,10 +90,14 @@ public static boolean validatePassword(Context context, String password, String
      */
     public static boolean validatePassword(Context context, String password) {
         if (password.length() < 8) {
-            Toast.makeText(context, "Password must be at least 8 characters!", Toast.LENGTH_SHORT).show();
+            Toast.makeText(context,
+                    R.string.password_short,
```
mccodav1: Suggest name like "error_password_short" so it is clear this is an error message of some type (as opposed to just a length)

app/src/main/java/com/example/poste/utils/ValidationUtils.java
```diff
@@ -76,10 +90,14 @@ public static boolean validatePassword(Context context, String password, String
...
+                    Toast.LENGTH_SHORT).show();
             return false;
         } else if (password.length() > 32) {
-            Toast.makeText(context, "Password must be at most 32 characters!", Toast.LENGTH_SHORT).show();
+            Toast.makeText(context,
+                    R.string.password_long,
```
mccodav1: Suggest name like "error_password_long" so it is clear this is an error message of some type (as opposed to just a length)

app/src/main/res/layout/activity_register.xml
```diff
@@ -28,7 +28,7 @@
         <EditText
             android:id="@+id/editTextEmail"
             android:layout_width="match_parent"
-            android:hint="Email Address"
+            android:hint="@string/email_address"
```
mccodav1: Suggest naming all these hints more descriptively; we want to ensure hints and fields are properly abstracted so we can keep them separate from other places that might use the text "Email Address" (to allow changing of hints and strings separately)

Suggest something like "hint_email_address"

mccodav1 (changes requested):

app/src/main/java/com/example/poste/activities/AccountActivity.java
atreichl: it already does.
image

app/src/main/java/com/example/poste/activities/NewPostActivity.java
atreichl: changed

app/src/main/java/com/example/poste/activities/AccountActivity.java
mccodav1: What I'm saying is, the above failure case says "Password change failed" but this one says "Edit failed". They do not match.

PR #48, opened 2023-11-28
User story 278: extend search functionality to tags

techbykim:


fmehdad97 (approved):
Tested and found to be working as expected.

In the future it might be nice to be able to search by leveraging multiple tags using OR and AND. Pending discussion on whether that's something the sponsor wants

PR #49, opened 2023-11-28
Abstract colors to colors.xml

fmehdad97:
Abstract out colors
Make drawables have transparentt background
Also added sponsors color pallette to colors.xml


atreichl (approved):
looks good I went though all the screens in the app and all the colors look correct.

PR #51, opened 2023-11-29
Project cleanup

mccodav1:
This branch cleans up a variety of things that have been bothering me for a while:

- Adds a TextView to Dashboard when no Folders exist. This will be marked as invisible until data is updated, and, if no folders exist, made visible. This occurs everytime data is updated.

- DashboardActivity and FolderViewActivity refreshing. Previously, these had a lot of overhead involved in recreating the activity, adapters, etc. every time data was updated. This has been cleaned up, so now the recyclerviews are properly updated when data is retrieved from the server. This allows the recyclerviews to perform their job of automatically refreshing what they display without requiring a force reload of the activity (which saves on memory and CPU cycles) and cuts down on duplicate API calls (previously, the activities would refresh data twice unnecessarily). 

- Cleans up some logging messages for better debugging

- Sets titles to invisible until data is refreshed. Previously, the 'Empty Folder', Folder title, etc. TextViews would display their default value momentarily until data was updated; now, they will be invisible until data is updated, ensuring we don't have that jarring display. 

- SharedFolderActivity previously called a method to send a user to the Dashboard; however, this meant we could go back to that page and leave us in an unexpected state. This has been changed to `finish()` that activity instead.

- Gradle updated (make sure to sync gradle: On Windows, Ctrl+Shift+O)

- Fixes a bug present on Dashboard where pressing back button the first time would refresh it, and a second press was necessary to minimize the app. This occurred because LoginActivity was finished, but IntroActivity wasn't. We don't want to always finish intro activity, as we might return to it from RegisterActivity; therefore, when a user successfully logs in, the application now clears the Activity stack, ensuring Dashboard is at the top level. 

- Fixes Dashboard, FolderView, and Search layout constraints. Previously, the constraints were improperly set so the display was off on screens that didn't exactly match that of the emulated Android Studio device. Constraints have been set such that they are dynamic according to screen size; this means that layouts should still appear correct when using different phones. I tested this on my physical device; previously elements scrolled off the screen or weren't aligned properly, and they are now, regardless of screen size.

This has been tested against dev server; advise if any issues encountered.

### DIFF
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -39,10 +39,8 @@ import retrofit2.Response;
  * The DashboardActivity class adds functionality to the activity_dashboard.xml layout
  */
 public class DashboardActivity extends PActivity {
-    private User currentUser;
     private Button addButton;
     private Button searchButton;
-    private List<com.example.poste.models.Folder> userFolders;
     private RecyclerView folderRecyclerView;
     private FolderAdapter folderAdapter;
     private ImageView optionsView;
@@ -64,12 +62,11 @@ public class DashboardActivity extends PActivity {
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
         configureWindow();
         prepVars();
         setListeners();
         createUpdateCallback();
-        currentUser.updateFoldersAndPosts(updateCallback);
+        User.getUser().updateFoldersAndPosts(updateCallback);
     }
 
     /**
@@ -98,37 +95,8 @@ public class DashboardActivity extends PActivity {
              */
             @Override
             public void onSuccess() {
-                userFolders = currentUser.getFolders(); // update our local copy of the folders
-                // Fill folder view (Recycler View)
-                // Note: This is not the best way to do this, because it means every time
-                // DashboardActivity restarts, a new Adapter is created, but it works for now. The
-                // better way would be to use DiffUtil to calculate the difference between the old
-                // and new lists, and then update the RecyclerView accordingly using
-                // AdapterListUpdateCallback. This method is an attempt to keep the code simple
-                // For more information, see:
-                // https://developer.android.com/reference/androidx/recyclerview/widget/DiffUtil
-                // and
-                // https://developer.android.com/reference/androidx/recyclerview/widget/ListUpdateCallback
-                folderAdapter = new FolderAdapter(
-                        new FolderAdapter.ClickListener() {
-                            @Override
-                            public void onItemClick(int position, Folder model) {
-                                // Send to that folder's view
-                                User.getUser().setSelectedFolder(model);
-                                Intent intent = new Intent(DashboardActivity.this, FolderViewActivity.class);
-                                startActivity(intent);
-                            }
-
-                            @Override
-                            public void onItemLongClick(int position, Folder model) {
-                                // Prompt to share folder
-                                Toast.makeText(DashboardActivity.this, getString(R.string.share_folder), Toast.LENGTH_LONG).show();
-                            }
-                        },
-                        new ArrayList<>(userFolders)
-                );
-                folderRecyclerView.setAdapter(folderAdapter);
-                registerForContextMenu(folderRecyclerView);
+                folderAdapter.setLocalDataSet(User.getUser().getFolders());
+                folderAdapter.notifyDataSetChanged();
             }
 
             /**
@@ -176,13 +144,32 @@ public class DashboardActivity extends PActivity {
      * initializes them. This saves us from resetting them every time the activity is restarted.
      */
     private void prepVars() {
-        currentUser = User.getUser(); // always get updated user, in case someone else logged in
         if (optionsView == null) {
             optionsView = findViewById(R.id.optionsbtn);
         }
         if (folderRecyclerView == null) {
             folderRecyclerView = findViewById(R.id.folder_recycler_view);
             folderRecyclerView.setLayoutManager(new GridLayoutManager(this, 4));
+            folderAdapter = new FolderAdapter(
+                    new FolderAdapter.ClickListener() {
+                        @Override
+                        public void onItemClick(int position, Folder model) {
+                            // Send to that folder's view
+                            User.getUser().setSelectedFolder(model);
+                            Intent intent = new Intent(DashboardActivity.this, FolderViewActivity.class);
+                            startActivity(intent);
+                        }
+
+                        @Override
+                        public void onItemLongClick(int position, Folder model) {
+                            // Prompt to share folder
+                            Toast.makeText(DashboardActivity.this, getString(R.string.share_folder), Toast.LENGTH_LONG).show();
+                        }
+                    },
+                    new ArrayList<>(User.getUser().getFolders())
+            );
+            folderRecyclerView.setAdapter(folderAdapter);
+            registerForContextMenu(folderRecyclerView);
         }
         if (addButton == null) {
             addButton = findViewById(R.id.dashboard_add_folder_btn);
@@ -277,7 +264,7 @@ public class DashboardActivity extends PActivity {
 
             // Handle folder creation logic
             Call<ResponseBody> call = apiService.createFolder(
-                    currentUser.getTokenHeaderString(),
+                    User.getUser().getTokenHeaderString(),
                     new FolderRequest(itemName)
             );
             call.enqueue(new Callback<ResponseBody>() {
@@ -351,7 +338,7 @@ public class DashboardActivity extends PActivity {
             case R.id.ctx_menu_delete_folder:
                 // Delete the folder
                 Call<ResponseBody> call = apiService.deleteFolder(
-                        currentUser.getTokenHeaderString(),
+                        User.getUser().getTokenHeaderString(),
                         folder.getId()
                 );
                 call.enqueue(new Callback<ResponseBody>() {
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -210,7 +210,6 @@ public class DashboardActivity extends PActivity {
         // Setup refresh
         if (swipeRefreshLayout == null) {
             swipeRefreshLayout = (SwipeRefreshLayout) findViewById(R.id.swipe_refresh_layout);
-            //swipeRefreshLayout.setOnRefreshListener(this::reloadDashboard);
             swipeRefreshLayout.setOnRefreshListener(() -> {
                 updateData();
                 swipeRefreshLayout.setRefreshing(false);
--- a/app/src/main/java/com/example/poste/activities/DashboardActivity.java
+++ b/app/src/main/java/com/example/poste/activities/DashboardActivity.java
@@ -51,7 +51,13 @@ public class DashboardActivity extends PActivity {
     @Override
     protected void onRestart() {
         super.onRestart();
-        reloadDashboard();
+        updateData();
+    }
+
+    private void updateData() {
+        // set any titles to invis
+        User.getUser().updateFoldersAndPosts(updateCallback);
+        // set any titles to vis
     }
 
     /**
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -98,37 +101,11 @@ private void createUpdateCallback() {
...
-                // Note: This is not the best way to do this, because it means every time
-                // DashboardActivity restarts, a new Adapter is created, but it works for now. The
-                // better way would be to use DiffUtil to calculate the difference between the old
-                // and new lists, and then update the RecyclerView accordingly using
-                // AdapterListUpdateCallback. This method is an attempt to keep the code simple
-                // For more information, see:
-                // https://developer.android.com/reference/androidx/recyclerview/widget/DiffUtil
-                // and
-                // https://developer.android.com/reference/androidx/recyclerview/widget/ListUpdateCallback
-                folderAdapter = new FolderAdapter(
```
mccodav1, line 112: Moving this to vars prepping, and simply updating the adapter and notifying of changes

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -98,37 +101,11 @@ private void createUpdateCallback() {
              */
             @Override
             public void onSuccess() {
-                userFolders = currentUser.getFolders(); // update our local copy of the folders
```
mccodav1, line 101: Rather use User.getUser() each time, rather than storing a user created by singleton

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -39,21 +39,25 @@
...
     
     @Override
     protected void onRestart() {
         super.onRestart();
-        reloadDashboard();
+        updateData();
+    }
+
+    private void updateData() {
+        dashboardViewEmptyText.setVisibility(View.GONE);
```
mccodav1, line 60: Sets the empty dashboard notification to invisible; the callback makes it visible if there are no folders (avoids the notification popping in and out)

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -98,37 +101,11 @@ private void createUpdateCallback() {
...
-                        },
-                        new ArrayList<>(userFolders)
-                );
-                folderRecyclerView.setAdapter(folderAdapter);
-                registerForContextMenu(folderRecyclerView);
+                folderAdapter.setLocalDataSet(User.getUser().getFolders());
+                folderAdapter.notifyDataSetChanged();
+                if (User.getUser().getFolders().size() == 0) {
+                    dashboardViewEmptyText.setVisibility(View.VISIBLE);
+                }
```
mccodav1, line 109: This enables the empty dash notification, if necessary. Avoids popping in and out.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -176,13 +156,36 @@ private void setListeners() {
...
+                        @Override
+                        public void onItemLongClick(int position, Folder model) {
+                            // Prompt to share folder
+                            Toast.makeText(DashboardActivity.this, getString(R.string.share_folder), Toast.LENGTH_LONG).show();
+                        }
+                    },
+                    new ArrayList<>(User.getUser().getFolders())
+            );
+            folderRecyclerView.setAdapter(folderAdapter);
+            registerForContextMenu(folderRecyclerView);
```
mccodav1, line 189: Moved adapter / recyclerview instantiation to this method so it doesn't get called over and over

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -207,23 +210,14 @@ private void configureWindow() {
         // Setup refresh
         if (swipeRefreshLayout == null) {
             swipeRefreshLayout = (SwipeRefreshLayout) findViewById(R.id.swipe_refresh_layout);
-            swipeRefreshLayout.setOnRefreshListener(this::reloadDashboard);
+            //swipeRefreshLayout.setOnRefreshListener(this::reloadDashboard);
+            swipeRefreshLayout.setOnRefreshListener(() -> {
+                updateData();
+                swipeRefreshLayout.setRefreshing(false);
+            });
```
mccodav1, line 217: Swipe refresh will now just update data; the recyclerview will update automatically without reloading activity.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
```diff
@@ -207,23 +210,14 @@ private void configureWindow() {
...
-     * This is intended to be used by the swipe refresh layout to allow the user to refresh the
-     * dashboard by swiping down
-     */
-    private void reloadDashboard() {
-        Intent intent = new Intent(DashboardActivity.this, DashboardActivity.class);
-        intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
-        startActivity(intent);
-        finish();
-        swipeRefreshLayout.setRefreshing(false);
-    }
```
mccodav1, line 225: No longer needed; recyclerview updates dynamically.

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -38,6 +42,9 @@ public class FolderViewActivity extends AppCompatActivity {
     private Button newBut;
     private ImageView settingsBut;
 
+    private SwipeRefreshLayout swipeRefreshLayout;
+
```
mccodav1, line 46: Added swipe refresh to FolderView

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -84,47 +91,56 @@ private void setListeners() {
 
     private void prepVars() {
         emptyText = findViewById(R.id.folderViewEmptyText);
+        emptyText.setVisibility(View.GONE); // Hide empty text until data is loaded
         folderName = findViewById(R.id.folderNameText);
+        folderName.setVisibility(View.INVISIBLE); // Hide folder name until data is loaded
+        // reset selected folder
```
mccodav1, line 97: Making these invisible by defaults avoids popin and popout of the placeholders; instead, they'll only be made visible if necessary after updating data

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -84,47 +91,56 @@ private void setListeners() {
...
+        User.getUser().setSelectedFolder(User.getUser().getFolder(User.getUser().getSelectedFolder().getId()));
```
mccodav1, line 98: This should be refactored at some point, but works for now

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -84,47 +91,56 @@ private void setListeners() {
...
+
+                    @Override
+                    public void onItemLongClick(int position, Post post) {
+                        User.getUser().setSelectedPost(post);
+                    }
+                },
+                User.getUser().getSelectedFolder().getPosts()
+        );
+        postRecyclerView.setAdapter(postAdapter);
+        postRecyclerView.setLayoutManager(new LinearLayoutManager(FolderViewActivity.this));
```
mccodav1, line 127: Moved all of this into prepVars for consistency (and to avoid calling it numerous times)

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -84,47 +91,56 @@ private void setListeners() {
...
-                // Set folder name in title bar
-                folderName.setText(User.getUser().getSelectedFolder().getTitle());
-
-                // Remove empty text if posts exist in folder or remove buttons if the folder is empty
+                postAdapter.setLocalDataSet(User.getUser().getSelectedFolder().getPosts());
+                postAdapter.notifyDataSetChanged();
+                List<Post> newPosts = User.getUser().getSelectedFolder().getPosts();
+                for (Post post : newPosts) {
+                    Log.d("FolderViewActivity", "Post: " + post.getTitle());
+                }
```
mccodav1, line 138: Update recyclerview, which refreshes dynamically, and set emptyText visible if necessary

app/src/main/java/com/example/poste/activities/FolderViewActivity.java
```diff
@@ -147,6 +163,15 @@ private void configureWindow() {
...
         // Set the activity layout
         setContentView(R.layout.activity_folder_view);
+
+        // Setup refresh
+        if (swipeRefreshLayout == null) {
+            swipeRefreshLayout = (SwipeRefreshLayout) findViewById(R.id.swipe_refresh_layout);
+            swipeRefreshLayout.setOnRefreshListener(()-> {
+                updateData();
+                swipeRefreshLayout.setRefreshing(false);
+            });
```
mccodav1, line 173: Updates data if swiping down

app/src/main/java/com/example/poste/adapters/FolderAdapter.java
```diff
@@ -26,6 +26,10 @@ public class FolderAdapter extends RecyclerView.Adapter<FolderAdapter.ViewHolder
     private List<Folder> localDataSet;
     public int position;
 
+    public void setLocalDataSet(List<Folder> userFolders) {
+        this.localDataSet = userFolders;
+    }
```
mccodav1, line 31: Allows us to update the adapter, rather than creating a brand new one each time

app/src/main/java/com/example/poste/models/Post.java
```diff
@@ -95,11 +95,7 @@ public String getUrl() {
     public String getId() { return id; }
 
     public ArrayList<String> getTags() {
-        ArrayList<String> ret = new ArrayList<>();
-        for (String s : tags) {
-            ret.add(s);
-        }
-        return ret;
+        return new ArrayList<>(tags);
```
mccodav1, line 98: Really, we could probably just return tags, but in sticking with the original intent of returning a copy, I just cleaned it up a bit. No need to parse a list into strings to create a new list

app/src/main/res/layout/activity_dashboard.xml
```diff
@@ -83,6 +83,21 @@
...
+                    android:textAlignment="center"
+                    android:textColor="#3B3B3B"
+                    android:textSize="25sp"
+                    app:layout_constraintBottom_toBottomOf="parent"
+                    app:layout_constraintEnd_toEndOf="parent"
+                    app:layout_constraintHorizontal_bias="0.5"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent"
+                    app:layout_constraintVertical_bias="0.5" />
+
```
mccodav1, line 44: Notification to add a Folder to dashboard when empty

app/src/main/res/layout/activity_folder_view.xml
```diff
@@ -1,113 +1,116 @@
 <?xml version="1.0" encoding="utf-8"?>
```
mccodav1, line 1: This looks like a big diff, but it only adds the swiperefresh. Everything else gets moved down one level and indented; nothing else changed.

atreichl:
I am unable to test it against the dev server I get this error.
image
testing local with the dev branch I do not have this issue.

mccodav1:
> I am unable to test it against the dev server I get this error. image testing local with the dev branch I do not have this issue.

That is a backend issue, I think the server had the wrong branch up. Try again.

app/src/main/java/com/example/poste/activities/SharedFolderActivity.java
```diff
@@ -104,15 +104,11 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
                 e.printStackTrace();
                 Toast.makeText(this, R.string.internal_error, Toast.LENGTH_LONG).show();
             }
-            sendToDashboard();
+            finish();
```
mccodav1, line 107: sendToDashboard() just started a new Dashboard activity, and didn't finish this one. A finish() is all that is necessary here

atreichl (commented):
If I shared a folder, once I am back on the dashboard if I then swipe to go back it takes me back to the sharefolder screen is this the intended behavior? Everything else looks good from my testing.

https://github.com/Appyo-Poste/Poste/assets/127570704/c203bfef-7404-4486-8079-1f1c988213b0

mccodav1:
> If I shared a folder, once I am back on the dashboard if I then swipe to go back it takes me back to the sharefolder screen is this the intended behavior? Everything else looks good from my testing.
> dashboard.back.to.share.folder.mp4

I see you are on the correct branch; try pulling again. That should not occur; I imagine you are missing some commits locally. When I run on this branch, pressing back will not take you back to the Share Folder activity

atreichl (approved):
Everything looks good from my testing. share folder works as expected going back to dashboard.

app/src/main/java/com/example/poste/activities/DashboardActivity.java
fmehdad97: Will it work on a failed to update also?

fmehdad97 (approved):
I like the changes!

app/src/main/java/com/example/poste/activities/DashboardActivity.java
mccodav1: Yes, as it is also set in the onError method of the callback

PR #53, opened 2024-01-10
Lock text entry of folder names, post names and links to be only 1 line

dzemlin:
Simply adds a property to all text fields for titles and links to allow only one line.
This makes a press of the enter key simply confirm the currently entered text instead of creating a new line.


fmehdad97 (approved):
Not really sure what has changed in the UX. But pressing ENTER on my keyboard takes me to the next field, on both fields that have the singleLine="true" and those that dont (such as Description). Approving since nothing is broken

PR #54, opened 2024-01-15
added * to fields that are required

atreichl:
added * to fields that are required on screens with fields that are not all required. Also added a message to said screens explaining what the * means.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -100,8 +100,8 @@
     <string name="create">Create</string>
     <string name="delete">Delete</string>
     <string name="edit">Edit</string>
-    <string name="folder">Folder</string>
-    <string name="folder_name">Folder Name</string>
+    <string name="folder">Folder*</string>
+    <string name="folder_name">Folder Name*</string>
     <string name="name">Name</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -100,8 +100,8 @@
     <string name="create">Create</string>
     <string name="delete">Delete</string>
     <string name="edit">Edit</string>
-    <string name="folder">Folder</string>
-    <string name="folder_name">Folder Name</string>
+    <string name="folder">Folder*</string>
```
fmehdad97: This has an unintended effect of showing the * in the dialog pop up when you click the + to add a folder or post

fmehdad97 (changes requested):
The message that says Fields with * are required.

However, I would request the following changes:
1) Changing the * to red color as is convention on most applications. There's more than one way of doing it. All of which are fairly easy to do. 
2) On the register menu, there is no indication of which fields are required (all are required, but on almost all applications it is still annotated). Adding a texxtview saying all fields required or adding the * to all fields should work.

fmehdad97 (approved):
Looks good!

PR #55, opened 2024-01-17
Abstract all drawables

fmehdad97:
All images used have been abstracted to drawables.xml
This allows you to add a new drawable object in the folder, and then simply replace the value that the abstract refers to, and it will change that in all the places in the application. 

One convention I tried to follow for ease of understanding, I named abstracts such as post_icon, folder_icon, poste_logo, and naming for the actual drawables were such as post_icon1, folder_icon0, etc. 

How to test: 
Change the abstract value in drawables.xml
Build and run the program
Make sure the asset changes in all the different places that it shows up


atreichl (approved):
The current symbols in the app are all correct, and when they are changed in drawable.xml they are changed though out the whole app.

PR #57, opened 2024-02-03
Changes dev branch to use compose prototype application

mccodav1:
Rather than working on a separate development branch, this makes the dev branch contain the Compose application we are building


fmehdad97 (approved):
I think moving to a dev with Compose is the right move since the dev branch:
a) The dev branch wouldn't be receiving any new updates
b) With the nested folder changes on the dev of backend, the dev branch would break anyways
c) Gives back all the niceties of CI/CD and branch protection

PR #59, opened 2024-02-12
Merged Dev into Composables Branch, and refactored objects

fmehdad97:
Refactored so we can both work on things. 

Top priority task will be to pull list of folders (shared or not) from server, use that data to populate the lazy column. 
Have to figure out what date does still (sponsor hasn't responded YET).


mccodav1 (approved):
Take a look at the new prototypes, which will change how things are represented (again... ugh). Also will need to make sure we get composables for Posts, too

PR #60, opened 2024-02-16
Complete intro screens (login, register) and redirect to a basic dashboard

mccodav1:


app/src/main/java/com/poste/DashboardScreen.kt
fmehdad97, line 1: I may have coded basicallly the exact same thing. However, mine are in the form of individual composables which can be pasted. I'll refactor them during my next sitting.

fmehdad97 (approved):

PR #61, opened 2024-02-25
Dashboard with Folderview, TopBar with Search Bar

fmehdad97:
Remaining:
Post grid. Post grid title and Folder colum titles

image


mccodav1:
We're duplicating efforts here; I implemented the same thing. I think it's imperative we get together ASAP to split up work to avoid

mccodav1 (approved):

PR #62, opened 2024-02-28
Login and Dashboard improvements

mccodav1:
- Tokens now properly received, parsed, and stored during login
- Refactors how API calls are made to follow better separation of concerns / MVVM pattern:
  - Introduction of Data repository
  - Introduction o Models
  - Introduction of Views / Viewmodels
This means that the data should be automatically marshaled into the format we need when receiving data from the backend on folders, posts, and shared folders. This should make it easier to work with moving forward. For now, this no longer uses sample data, and instead uses data sent from the backend (provided the backend sends the requisite data; see the Model classes for what it looks for / defaulst to)


dzemlin (approved):
based purely on observation I approve this. It appears to run from the video and i don't see any code issues that jump out at me. But i would like to discuss this part of the app more so i can be more helpful in future pull request.

fmehdad97 (approved):

PR #64, opened 2024-03-13
Add create folder dialog and attach to button

mccodav1:


fmehdad97 (approved):
A better and more complete implementation than what I had.

PR #71, opened 2024-04-10
Fix Folder model parsing of Backend data

mccodav1:
Because the backend has changed in how it returns folder data (nested Folders) the Folder model needed to be changed to avoid application crashes. 

This branches off of and merges to Fayez's `fmehdad/Sprint6/CreateAndEditPosts` branch.


fmehdad97:
Thanks for the hotfix, I wasn't aware my folder parsing broke

PR #74, opened 2024-04-17
Enable folder traversal

mccodav1:
This enables folder traversal by navigating to a new Dashboard instance for each folder, popping them off the stack as the user goes back. This means that each Dashboard and sub-elements has their own viewmodel with shared data; the data updates as we navigate forwards and backwards as expected.

Next to do: We should forward the current folder's name (if not root) to the topbar so we can present the folder title and a back button; this topbar element can be null by default, and filled in only if data is provided. In each folder, the dashboard has the folderId variable containing the folder's id: at the root level, the folderId is 'root', so that helps.

We also should consider passing the folderId (if NOT root) to post / folder creation and using that as the default creation location (so a new folder is created inside the folder the user is in, etc.)


fmehdad97:
I agree with your comments. Passing the folder Id to folder creation is really easy right now. However, I previously ran into a lot of trouble when trying to pass data into the screen. It's solvable for sure (I could get the data, but it would crash immediately), but with the dire lack of time remaining I chose not to stay stuck on that for now and tackle other problems.

fmehdad97 (approved):

fmehdad97:
PR approved. 
However, a bug found : all folders created end up in "root". 
@mccodav1 has confirmed it is expected behavior (for now), and the feature has not been implemented yet.

